PR #26, opened 2018-07-26
Pyzhou/tdi 40711 migrate mongo db tacokit

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-40711


wwang-talend (dismissed):
LGTM, don't forget migration

pyzhou-talend:
Only for the usage of web , so there's  no need for migration for now.

PR #29, opened 2018-08-16
feat: Move CustomData code from TCOMPv0.

RyanSkraba:
Extract code for the LocalIO component from a POC branch for master.


localio/src/main/java/org/talend/components/localio/fixed/FixedDataSetConfiguration.java
```diff
@@ -20,17 +17,16 @@
...
-public class FixedDataSet implements Serializable {
+public class FixedDataSetConfiguration implements Serializable {
 
     @Option
-    private FixedDatastore datastore;
+    private FixedDataStoreConfiguration datastore = new FixedDataStoreConfiguration();
 
     @Option
-    @Required
+    // @Required
```
pbailly: Dead code.

PR #43, opened 2018-09-27
WIP TDI-40707 - Jdbc :: migrate cloud component to tacokit (Data store/Sata set) + Input/Output

akhabali:
https://jira.talendforge.org/browse/TDI-40707


jdbc/src/main/java/org/talend/components/jdbc/dataset/OutputDataset.java
```diff
@@ -0,0 +1,83 @@
...
+    @Documentation("")
+    private List<UpdateOperationMapping> updateOperationMapping;
+
+    @Option
+    @ActiveIf(target = "actionOnData", value = "Delete")
+    @Documentation("")
+    private List<String> deleteKeys;
+
+    public enum ActionOnData {
+        Insert,
```
mbasiuk-talend: Should we use uppercase for enum elements ? (it might be a convetion question)
And for display name we can use this version of naming.

jdbc/src/main/java/org/talend/components/jdbc/output/Output.java
```diff
@@ -0,0 +1,171 @@
...
+            } catch (SQLException e1) {
+                log.error("Can't rollback statements", e);
+            }
+
+            // fixme : should we transform this component to a processor and :
+            // 2 . emit rejected records
+
+            StringBuilder batchErrorMessage = new StringBuilder();
+            SQLException batchError = e;
+            while (batchError.getNextException() != null) {
```
mbasiuk-talend: Will you miss first exception in a batch here?

jdbc/src/main/java/org/talend/components/jdbc/output/Output.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    @PreDestroy
+    public void preDestroy() {
+        if (preparedStatement != null) {
+            try {
+                preparedStatement.close();
+            } catch (SQLException e) {
+                log.warn("Can't close the prepared statement properly.", e);
```
mbasiuk-talend: i18n?

jdbc/src/main/java/org/talend/components/jdbc/output/Output.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+
+        if (connection != null) {
+            try {
+                connection.close();
+            } catch (SQLException e) {
+                log.warn("Can't close the connection properly.", e);
```
mbasiuk-talend: i18n?

jdbc/src/main/java/org/talend/components/jdbc/output/internal/RecordSQLTypes.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+package org.talend.components.jdbc.output.internal;
+
+import java.sql.Date;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+import org.talend.sdk.component.api.record.Record;
+
+public enum RecordSQLTypes {
```
mbasiuk-talend: Do we support set **NULL**?

jdbc/src/main/java/org/talend/components/jdbc/service/ActionService.java
```diff
@@ -1,76 +1,119 @@
...
-            throw new IllegalArgumentException(i18n.errorEmptyJdbcURL());
-        }
-        final URLClassLoader loader = jdbcDriversService.getDriverClassLoader(datastore.getDbType());
-        if (loader == null) {
-            return new HealthCheckStatus(HealthCheckStatus.Status.KO, i18n.errorCantLoadDriver(datastore.getDbType()));
-        }
         try {
-            Driver driver = Driver.class
-                    .cast(loader.loadClass(jdbcDriversService.getDrivers().get(datastore.getDbType()).getClazz()).newInstance());
+            this.jdbcDriversService.connection(datastore);
```
mbasiuk-talend: I think we should close connection here, or not close it in getTableFromDatabase in try-with-resources block. Which is correct here?

jdbc/src/main/resources/TALEND-INF/local-configuration.properties
```diff
@@ -0,0 +1,32 @@
...
+#  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+#  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+#  specific language governing permissions and limitations under the License.
+#
+# MySQL
+#
+jdbc.drivers[0].id=MySQL
+jdbc.drivers[0].className=com.mysql.jdbc.Driver
+jdbc.drivers[0].url=jdbc:mysql://localhost:3306/db
+jdbc.drivers[0].paths[0].path=mysql:mysql-connector-java:5.1.38
```
mbasiuk-talend: Are we going to use still very old driver? **Dec, 2015**

jdbc/src/main/java/org/talend/components/jdbc/dataset/OutputDataset.java
akhabali: i'm not aware of any naming conventions for enums. and we have display names in i18n resources where we can write human readable words.

jdbc/src/main/java/org/talend/components/jdbc/dataset/OutputDataset.java
```diff
@@ -0,0 +1,83 @@
...
+    @Data
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @AutoLayout
+    public static class UpdateOperationMapping {
+
+        // fixme : use the values from schema when available in tacokit
+        @Option
+        @Documentation("")
+        private String column;
```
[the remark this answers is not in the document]
akhabali: those columns need to get gathered from the schema. this is something we don't support yet in tacokit. see my comment above.

jdbc/src/main/java/org/talend/components/jdbc/output/Output.java
akhabali: it's on purpose. the first message just tell you that you have errors. and the full exception is kept as a cause.

jdbc/src/main/java/org/talend/components/jdbc/service/ActionService.java
akhabali: nice catch 👍

jdbc/src/main/java/org/talend/components/jdbc/output/internal/RecordSQLTypes.java
akhabali: Normally we need to support null values in the component.

i need to add more unit tests to cover this part. Also i need to align the statement parameters initialization with the new Record API which remove the null fields from the Entries list.

jdbc/src/main/java/org/talend/components/jdbc/input/InputEmitter.java
```diff
@@ -38,102 +45,68 @@
...
-                throw new IllegalStateException(e);
+            } else {
+                statement = connection.createStatement();
             }
-        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-            throw new IllegalStateException(i18n.errorCantLoadDriver(dbType));
+            resultSet = statement.executeQuery(query);
         } catch (SQLException e) {
-            throw new IllegalStateException(i18n.errorSQL(e.getErrorCode(), e.getMessage()));
+            throw new IllegalStateException(e);
```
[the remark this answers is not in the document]
akhabali: i d'ont have a custom error message here.

jdbc/src/main/java/org/talend/components/jdbc/dataset/OutputDataset.java
akhabali: i will use uppercase to be homogeneous with RecordSQLTypes enum below

PR #50, opened 2018-10-23
feat(TFD-5486): Rewrite Window processor

amaumont:
migrate Window processor from TCOMPV0 to Tacokit with unit tests


processing-beam/src/main/java/org/talend/components/processing/window/WindowConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+    @Max(Integer.MAX_VALUE)
+    @Documentation("The window duration (Data during X ms)")
+    private Integer windowLength = 5000;
+
+    @Option
+    @Required
+    @Min(0)
+    @Max(Integer.MAX_VALUE)
+    @Documentation("The window slide length (Every X ms)")
+    private Integer windowSlideLength = 5000;
```
bchen-talend: need to hidden when session checked
```suggestion
    @ActiveIf(target = "windowSession", value = "false")
    private Integer windowSlideLength = 5000;
```

processing-beam/src/main/resources/org/talend/components/processing/window/Messages.properties
```diff
@@ -0,0 +1 @@
+Processing.Window._displayName = Window
```
bchen-talend: better to specify _displayName for each property, else UI will show the var name.
WindowConfiguration.windowLength._displayName
WindowConfiguration.windowSlideLength._displayName
WindowConfiguration.windowSession._displayName

bchen-talend:
I didn't check jira description before the above comment, so you may waiting for answer about the new `Window type` and other property name.

processing-beam/src/main/java/org/talend/components/processing/window/WindowConfiguration.java
amaumont: Good point.

amaumont:
Required to create pull request from 'dss/tacokit_integration' not from 'master'

PR #51, opened 2018-10-25
fix(TFD-5339) : fix flatten migration bugs

amarouni:
- Remove technical labels
- Add appropriate doc
- Fix discard trailing strings functionality (iso Studio)

### DIFF
--- a/processing-beam/src/main/java/org/talend/components/processing/flatten/FlattenConfiguration.java
+++ b/processing-beam/src/main/java/org/talend/components/processing/flatten/FlattenConfiguration.java
@@ -20,7 +20,7 @@ public class FlattenConfiguration implements Serializable {
     @Option
     @Required
     @Documentation("")
-    private String columnToFlatten;
+    private String columnToFlatten = "";
 
     @Option
     @Required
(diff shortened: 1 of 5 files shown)


processing-beam/src/main/java/org/talend/components/processing/flatten/FlattenConfiguration.java
```diff
@@ -13,43 +12,41 @@
...
-
 @Data
 @Documentation("Flatten Configuration")
 @OptionsOrder({ "columnToFlatten", "isList", "fieldDelimiter", "specificFieldDelimiter", "discardTrailingEmptyStr", "trim" })
 public class FlattenConfiguration implements Serializable {
 
     @Option
     @Required
-    @Documentation("The column to flatten")
+    @Documentation("")
```
pbailly: Please add the documentation.

processing-beam/src/main/java/org/talend/components/processing/flatten/FlattenDoFn.java
```diff
@@ -21,7 +20,8 @@ public void processElement(ProcessContext context) throws Exception {
         boolean isDiscardTrailingEmptyStr = properties.isDiscardTrailingEmptyStr();
         boolean isTrim = properties.isTrim();
 
-        if (StringUtils.isNotEmpty(columnToFlatten)) {
+        if (!FlattenUtils.isNullOrEmpty(columnToFlatten)) {
```
pbailly: Please, avoid using negative boolean

PR #53, opened 2018-10-29
fix : rename flatten to normalize

amarouni:
Rename flatten to normalize


processing-tck/src/main/java/org/talend/components/processing/normalize/Normalize.java
```diff
@@ -23,6 +9,19 @@
 import org.talend.sdk.component.api.processor.OutputEmitter;
 import org.talend.sdk.component.api.processor.Processor;
 
+import javax.json.JsonBuilderFactory;
```
RyanSkraba: plz set your import order to the Talend standard!

PR #73, opened 2018-12-21
TDI-41448 - JDBC Snowflake 

akhabali:
https://jira.talendforge.org/browse/TDI-41448


akhabali:
replaced by https://github.com/Talend/connectors-se/pull/75

PR #78, opened 2019-01-07
TDI-41569 : Support --component-jdbc-auto-download-drivers for jdbc d…

akhabali:
…river loading in setup.jar

https://jira.talendforge.org/browse/TDI-41569


connectors-se-docker/src/main/java/org/talend/components/docker/ServerSetup.java
```diff
@@ -1,36 +1,24 @@
...
+import lombok.NoArgsConstructor;
 import org.talend.sdk.component.dependencies.maven.Artifact;
 import org.talend.sdk.component.dependencies.maven.MvnCoordinateToFileConverter;
 import org.tomitribe.crest.Main;
-import org.tomitribe.crest.api.Command;
-import org.tomitribe.crest.api.Default;
-import org.tomitribe.crest.api.Err;
-import org.tomitribe.crest.api.Option;
-import org.tomitribe.crest.api.Out;
+import org.tomitribe.crest.api.*;
```
rmannibucau: idea config issue?

connectors-se-docker/src/main/java/org/talend/components/docker/ServerSetup.java
```diff
@@ -1,36 +1,24 @@
...
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.PrintStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
-import java.nio.file.Files;
-import java.util.List;
-
-import javax.json.Json;
```
rmannibucau: if not needed you can drop the deps?

PR #91, opened 2019-02-07
fix(TFD-6784): Set Salesforce icon.

pbailly:
Set the correct salesforce icon.


salesforce/src/main/java/org/talend/components/salesforce/package-info.java
```diff
@@ -1,5 +1,5 @@
 @Components(family = "Salesforce", categories = "Business")
-@Icon(value = Icon.IconType.CUSTOM, custom = "salesforce")
+@Icon(Icon.IconType.FILE_SALESFORCE)
```
rmannibucau: you should delete the icon as well (from https://github.com/Talend/connectors-se/tree/master/salesforce/src/main/resources/icons)

pbailly: Everything?

rmannibucau: only the one you make "dead" with this PR so <customname>_icon32.png (saleforce_icon32.png)

PR #109, opened 2019-04-02
i18n(TDI-41939) : export (upload) i18n package to xtm

akhabali:


gonchar-ivan:
just interesting
What is purpose of this image? 
'khabali/jenkins-java-build-container:latest'

akhabali:
This image provides the build environment and contains mvn, java, docker and other required tools for the build like rsync.

The old image was a 32bit image and nodejs doesn't exist in 32 bit for Linux.

Also I wasn't able to found any existing image that already has all those tools

Jenkinsfile
```diff
@@ -145,6 +157,73 @@ spec:
...
+        stage('Push to Xtm') {
+            when {
+                anyOf{
+                    expression { params.PUSH_TO_XTM == true }
+//                    allOf{
+//                        triggeredBy 'TimerTrigger'
+//                        expression {
+//                            (calendar.get(Calendar.WEEK_OF_MONTH) == 2 ||  calendar.get(Calendar.WEEK_OF_MONTH) == 4) && calendar.get(Calendar.DAY_OF_WEEK) == Calendar.THURSDAY
+//                        }
+//                    }
```
gonchar-ivan: Don't forget to remove commented code someday ;-)

gonchar-ivan:
Thanks for explanation about image.
Also could you, please, tell a bit about json files: 
talend-i18n.json
package.json

Jenkinsfile
akhabali: this part automate the export. it will be activated soon ;)

akhabali:
It's all explained in the https://github.com/Talend/i18n-product project.
There is a node script that manage the export / import of i18n resources. we only do an integration of that in this PR

PR #121, opened 2019-04-12
feat(TDI-40612): Marketo People and Activities

undx:
https://jira.talendforge.org/browse/TDI-40708


marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
```diff
@@ -0,0 +1,143 @@
...
+    public JsonObject runAction() {
+        switch (configuration.getLeadAction()) {
+        case getLead:
+            return getLead();
+        case getMultipleLeads:
+            return getMultipleLeads();
+        case getLeadActivity:
+            return getLeadActivities();
+        case getLeadChanges:
+            return getLeadChanges();
```
chmyga: Could be nice to implement a Strategy here not to run through this switch case during every `runAction()` call.

PR #134, opened 2019-05-15
feat(TDI-40639): N1QL integration

Voievidko:
https://jira.talendforge.org/browse/TDI-40639


couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
```diff
@@ -0,0 +1,223 @@
...
+            throw new CouchbaseException(i18n.connectionKO());
+        }
+
+        bucket.bucketManager().createN1qlPrimaryIndex(true, false);
+
+        columnsSet = new HashSet<>();
+
+        N1qlQueryResult n1qlQueryRows;
+        if (configuration.isUseN1QLQuery()) {
+            n1qlQueryRows = bucket.query(N1qlQuery.simple(configuration.getQuery()));
```
ypiel-talend: Did you ask if you should use simple n1ql or parametrized ones ? https://docs.couchbase.com/sdk-api/couchbase-java-client-2.6.2/com/couchbase/client/java/query/N1qlQuery.html#simple-java.lang.String-com.couchbase.client.java.query.N1qlParams-

it could be more safe and prevent from code injection... but it implies to add an array with key/value to set params.

couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
```diff
@@ -0,0 +1,223 @@
...
+            n1qlQueryRows = bucket.query(N1qlQuery.simple("SELECT * FROM `" + bucket.name() + "`"));
+        }
+        checkErrors(n1qlQueryRows);
+        index = n1qlQueryRows.rows();
+    }
+
+    private void checkErrors(N1qlQueryResult n1qlQueryRows) {
+        if (!n1qlQueryRows.errors().isEmpty()) {
+            LOG.error(i18n.queryResultError());
+            throw new IllegalArgumentException(n1qlQueryRows.errors().toString());
```
ypiel-talend: You call toString on a List<JsonObject> ? Should not be readable...
https://docs.couchbase.com/sdk-api/couchbase-java-client-2.3.6/com/couchbase/client/java/query/N1qlQueryResult.html#errors--

couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private Schema parseSchema(JsonObject document) {
+        final Schema.Builder schemaBuilder = builderFactory.newSchemaBuilder(RECORD);
+        columnsSet.stream().forEach(name -> addField(schemaBuilder, name, getValue(name, document)));
+        return schemaBuilder.build();
+    }
+
+    private void addField(final Schema.Builder schemaBuilder, final String name, Object value) {
+        if (value == null) {
+            LOG.warn(i18n.cannotGuessWhenDataIsNull());
```
ypiel-talend: What is the type so ? Should not be your default in getSchemaType, STRING ?
Maybe better to deal with null in getSchemaType (to think about, just a proposal without knowing the technology)

couchbase/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1 @@
+distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.6.0/apache-maven-3.6.0-bin.zip
```
ypiel-talend: To remove I think

couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
Voievidko: For now it will be simple. We discuss it with Pierre

couchbase/wrapper/maven-wrapper.properties
Voievidko: Done

couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
Voievidko: My code try to guess type on the first value.
In my opinion we can not guess type if value is `null`.
And assign type to String when value is null can play bad game with us.
For example, suppose we have income JSON:
image
There are two JSON objects (Records)
By the first element we can't guess that Object `tel` is another JSON (Record), not actually the String.

couchbase/src/main/java/org/talend/components/couchbase/source/CouchbaseInput.java
Voievidko: Why not readable?
Try to run code sample on Intellij, looks ok
image

PR #148, opened 2019-06-04
Dsylaiev/tdi 42439 fix avro null values

dmytro-sylaiev:


azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -90,30 +108,53 @@ protected GenericRecord recordToAvro(Record fromRecord, GenericRecord toRecord)
...
             case LONG:
-                toRecord.put(name, fromRecord.getLong(name));
+                OptionalLong optionalLongValue = fromRecord.getOptionalLong(name);
+                if (optionalLongValue.isPresent()) {
+                    toRecord.put(name, optionalLongValue.getAsLong());
+                } else {
+                    toRecord.put(name, null);
+                }
                 break;
             case FLOAT:
```
undx: floats are managed like doubles, this could lead to potential errors and class cast exceptions.
I think that a `tacokitRecord.getFloat("field");`  won't work...
I recommend to set a real float value:
```java
toRecord.put(name, Double.valueOf(optionalFloat.getAsDouble()).floatValue());
```

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -217,52 +264,50 @@ protected Record avroToRecord(GenericRecord genericRecord, List<org.apache.avro.
...
-            if (field.schema().getType().equals(ARRAY)) {
+            Entry entry = inferAvroField(field);
+            if (org.apache.avro.Schema.Type.ARRAY.equals(field.schema().getType())) {
                 buildArrayField(field, value, recordBuilder, entry);
-
             } else {
-                buildField(field, value, recordBuilder, entry);
+                if (!entry.isNullable() || value != null) {
+                    buildField(field, value, recordBuilder, entry);
+                } // else skip null value for nullable field
```
undx: Skipping null values may lead to issues in pipelines.
TDP use first record to get its schema, if we've null value the field won't exist.

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -90,30 +108,53 @@ protected GenericRecord recordToAvro(Record fromRecord, GenericRecord toRecord)
...
+                }
                 break;
             case BOOLEAN:
-                toRecord.put(name, fromRecord.getBoolean(name));
+                Optional<Boolean> optionalBooleanValue = fromRecord.getOptionalBoolean(name);
+                if (optionalBooleanValue.isPresent()) {
+                    toRecord.put(name, optionalBooleanValue.get());
+                } else {
+                    toRecord.put(name, null);
+                }
```
chmyga: Better to do 
`toRecord.put(optionalBooleanValue.orElse(null));`

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -274,14 +319,14 @@ protected Record avroToRecord(GenericRecord genericRecord, List<org.apache.avro.
...
+        Schema.Builder builder = recordBuilderFactory.newSchemaBuilder(Type.RECORD);
         field.schema().getFields().stream().map(this::inferAvroField).forEach(builder::withEntry);
         return builder.build();
     }
 
     protected Schema buildArrayFieldSchema(org.apache.avro.Schema.Field field) {
-        Schema.Builder schemaBuilder = recordBuilderFactory.newSchemaBuilder(Schema.Type.RECORD);
-        Schema.Entry.Builder entryBuilder = recordBuilderFactory.newEntryBuilder();
+        Schema.Builder schemaBuilder = recordBuilderFactory.newSchemaBuilder(Type.RECORD);
+        Entry.Builder entryBuilder = recordBuilderFactory.newEntryBuilder();
```
chmyga: Arrays might contain simple objects, not Records.

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -344,52 +388,66 @@ protected void buildArrayField(org.apache.avro.Schema.Field field, Object value,
...
-            Schema.Entry entry) {
-        String logicalType = field.schema().getProp("logicalType"); // FIXME field.schema().getLogicalType() returns null
-        switch (field.schema().getType()) {
+    protected void buildField(org.apache.avro.Schema.Field field, Object value, Record.Builder recordBuilder, Entry entry) {
+        String logicalType = field.schema().getProp(AVRO_LOGICAL_TYPE);
+        org.apache.avro.Schema.Type fieldType = getFieldType(field);
+        switch (fieldType) {
         case RECORD:
-            recordBuilder.withRecord(entry, avroToRecord(GenericData.Record.class.cast(value), field.schema().getFields()));
+            recordBuilder.withRecord(entry, (Record) value);
```
chmyga: How will you get a component api Record from AvroRecord? I guess we should convert Record from AvroRecord as was before.

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
```diff
@@ -344,52 +388,66 @@ protected void buildArrayField(org.apache.avro.Schema.Field field, Object value,
...
+            byte[] bytes = value != null ? ((java.nio.ByteBuffer) value).array() : null;
+            recordBuilder.withBytes(entry, bytes);
             break;
         case INT:
-            if ("time-millis".equals(logicalType) || "date".equals(logicalType)) {
-                recordBuilder.withDateTime(entry,
-                        ZonedDateTime.ofInstant(Instant.ofEpochSecond((Integer) value), ZoneId.systemDefault()));
+            int ivalue = value != null ? (Integer) value : 0;
+            if (AVRO_LOGICAL_TYPE_DATE.equals(logicalType) || AVRO_LOGICAL_TYPE_TIME_MILLIS.equals(logicalType)) {
+                recordBuilder.withDateTime(entry, ZonedDateTime.ofInstant(Instant.ofEpochSecond(ivalue), ZoneOffset.UTC));
```
chmyga: if it's millis, how can you use epochsecond?

azure/azureblob/src/main/java/org/talend/components/azure/runtime/converters/AvroConverter.java
dmytro-sylaiev: Don't agree, it's OK with tests and only for nullable fields

PR #149, opened 2019-06-06
Dsylaiev/tdi 42460 create dropdown for action on output

dmytro-sylaiev:


azure/azureblob/src/main/java/org/talend/components/azure/output/BlobOutputConfiguration.java
```diff
@@ -33,6 +33,13 @@
     @Documentation("Azure Blob connection")
     private AzureBlobDataset dataset;
 
+    @Option
+    @Documentation("Output behavior for case when directory already exist.\n"
+            + "Default - throw an exception and stop component\n"
+            + "Append - new files will created in same directory (or appended for csv file). Should be used with same file format only to avoid data structure break\n"
+            + "Overwrite - delete all files from directory and insert new files.")
```
chmyga: `@DefaultValue("DEFAULT")` can be used here.

azure/azureblob/src/test/java/org/talend/components/azure/BlobTestUtils.java
```diff
@@ -81,11 +81,16 @@ public static void createStorage(String storageName, CloudStorageAccount connect
...
+    }
+
+    public static void uploadTestFile(CloudStorageAccount storageAccount, AzureBlobDataset dataset,
+                                      String resourceName, String targetName) throws URISyntaxException, StorageException, IOException {
         CloudBlobContainer container = storageAccount.createCloudBlobClient()
-                .getContainerReference(blobInputProperties.getDataset().getContainerName());
+                .getContainerReference(dataset.getContainerName());
         CloudBlockBlob blockBlob = container
-                .getBlockBlobReference(blobInputProperties.getDataset().getDirectory() + "/" + targetName);
+                .getBlockBlobReference(dataset.getDirectory() + "/" + targetName);
```
chmyga: What if directory ends with "/", wouldn't there be an error?

PR #156, opened 2019-06-11
feat(TDI-40635): Prepare for integration

chmyga:
* Fix error message on check connection

* Add family icon

* remove component runtime and api versions


couchbase/src/main/java/org/talend/components/couchbase/service/CouchbaseService.java
```diff
@@ -99,7 +101,16 @@ public HealthCheckStatus healthCheck(@Option("configuration.dataset.connection")
             openConnection(datastore);
             return new HealthCheckStatus(HealthCheckStatus.Status.OK, "Connection OK");
         } catch (Exception exception) {
-            return new HealthCheckStatus(HealthCheckStatus.Status.KO, exception.getMessage());
+            String message = "";
```
akhabali: can you log the exception please

PR #163, opened 2019-06-18
fix(TDI-42439): handling avro nulls

dmytro-sylaiev:


undx:
Hi Dmytro, see Jira issue updates.

PR #164, opened 2019-06-18
fix(TDI-42427): fix schemas, formats and converters

undx:
- correct records in record schema and record converter
- may also help/fix TDI-42439


undx:
Jira issue postponed
https://jira.talendforge.org/browse/TDI-42527

PR #166, opened 2019-06-19
fix(TDI-42525): set correct custom encoding when writing csv output blob

undx:
https://jira.talendforge.org/browse/TDI-42525


dmytro-sylaiev:
Please fix failing unit test for avroConverter

PR #169, opened 2019-06-24
fix(TDI-42547): fix schema for sub-record in AVRO and Parquet

undx:
https://jira.talendforge.org/browse/TDI-42547
https://jira.talendforge.org/browse/TDI-42551


azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/avro/AvroConverter.java
```diff
@@ -109,15 +112,26 @@ protected GenericRecord recordToAvro(Record fromRecord, GenericRecord toRecord)
         for (org.apache.avro.Schema.Field f : toRecord.getSchema().getFields()) {
             String name = f.name();
             org.apache.avro.Schema.Type fieldType = getFieldType(f);
+            org.apache.avro.Schema subSchema;
```
dmytro-sylaiev: just fyi you can declare variable in the first `case`, and then reuse, java allows it

azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/avro/AvroConverter.java
```diff
@@ -109,15 +112,26 @@ protected GenericRecord recordToAvro(Record fromRecord, GenericRecord toRecord)
...
+                Object firstArrayValue = new Object();
+                if (recordArray.iterator().hasNext()) {
+                    firstArrayValue = recordArray.iterator().next();
+                }
+                if (firstArrayValue instanceof Record) {
+                    subSchema = inferAvroSchema(((Record) firstArrayValue).getSchema());
+                    List<GenericRecord> records = recordArray.stream()
+                            .map(o -> recordToAvro((Record) o, new GenericData.Record(subSchema))).collect(Collectors.toList());
+                    toRecord.put(name, records);
+                } else {
```
dmytro-sylaiev: You'll put java.lang.Object if recordArray is initially empty, it it OK?

azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/avro/AvroConverter.java
undx: ah yes, thx

PR #181, opened 2019-07-09
chore(TFD-8167): give blank for placeholder of username/password/handler on JDBC

bchen-talend:
refer to https://jira.talendforge.org/browse/TFD-8167


bchen-talend:
change to https://github.com/Talend/connectors-se/pull/182

PR #188, opened 2019-07-16
fix(TDI-42623) : Required field issue for Connection and Dataset

undx:
https://jira.talendforge.org/browse/TDI-42623


undx:
Cancelled, will be fixed by tck fixes

PR #197, opened 2019-08-07
fix(TDI-42795): handle blobName with space(s)

undx:
https://jira.talendforge.org/browse/TDI-42803


mbasiuk-talend:
I would suggest to write unit tests for those specific conditions. Everything else is good for me.

PR #208, opened 2019-09-09
feat(TDI-42368): N1QL Parameterized Queries for CouchbaseOutput

undx:
https://jira.talendforge.org/browse/TDI-42368


couchbase/src/main/java/org/talend/components/couchbase/output/CouchbaseOutputConfiguration.java
```diff
@@ -37,7 +43,24 @@
...
-    @Required
     @Documentation("Field to use as ID")
+    @ActiveIf(target = "useN1QLQuery", value = "false")
     private String idFieldName;
-}
\ No newline at end of file
+
+    @Option
+    @Documentation("Use N1QL query")
+    private boolean useN1QLQuery = false;
```
mbasiuk-talend: I think this property should be before idFieldName. It's more common practice to hide below fields, not above.

couchbase/src/main/java/org/talend/components/couchbase/output/N1QLQueryParameter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@GridLayouts({ @GridLayout({ @GridLayout.Row({ "queryParameterName" }), @GridLayout.Row({ "column" }) }) })
+@Documentation("Couchbase input configuration")
```
mbasiuk-talend: Copy-paste?

couchbase/src/test/java/org/talend/components/couchbase/CouchbaseUtilTest.java
```diff
@@ -51,10 +52,13 @@
     public static final CouchbaseContainer COUCHBASE_CONTAINER;
 
     @Service
-    private RecordBuilderFactory recordBuilderFactory;
+    protected RecordBuilderFactory recordBuilderFactory;
+
+    public static final String COUCHBASE_SERVER_DOCKER_IMAGE = "couchbase/server:6.5.0-beta";
```
mbasiuk-talend: Are we able to use system properties here and change it in pom.xml? It would be more configurable.

couchbase/src/main/java/org/talend/components/couchbase/output/N1QLQueryParameter.java
undx: Oh, good catch ! Yeah a copy-paste from the Output config which was, I supposed, a copy-paste from Input config :smile: 
Will fix both

PR #213, opened 2019-09-17
fix(TDI-42967): handle other endpoint suffix

undx:
https://jira.talendforge.org/browse/TDI-42967


azure-dls-gen2/src/test/java/org/talend/components/adlsgen2/datastore/AdlsGen2ConnectionTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package org.talend.components.adlsgen2.datastore;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
mbasiuk-talend: it would be better to import only used assertions, not the whole class.

undx: blame my IDE ! :rofl: fixed !

PR #218, opened 2019-10-14
feat(TFD-7630)-triggers for datastores datasets

laurentgir:


component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
```diff
@@ -0,0 +1,99 @@
...
+                if (!clazz.isAnnotationPresent(DataStore.class)) {
+                    log.error("Invalid DatastoreRef.",
+                            new IllegalStateException("A @DatastoreRef can only be used on @Datastore parameters"));
+                } else {
+                    final String name = ((DataStore) clazz.getAnnotation(DataStore.class)).value();
+                    final String family = c.family();
+                    final HashMap<String, String> metas = new HashMap<>();
+                    metas.put(META_PREFIX + "family", family);
+                    metas.put(META_PREFIX + "name", name);
+                    metas.put(META_PREFIX + "type", "datastore");
```
laurentgir: Datastore hardcoded ?

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
```diff
@@ -0,0 +1,99 @@
...
+                    final DatastoreRef dsRef = (DatastoreRef) annotation;
+                    final AtomicInteger index = new AtomicInteger(0);
+                    Stream.of(dsRef.filters()).forEach(f -> {
```
laurentgir: Have we tested it ? 😆 
Does TCKProxy handle it ?

connections/connection-s3/src/main/java/org/talend/components/fileio/s3/S3Service.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+        return credentials;
+    }
+
+    public static AmazonS3 createClient(S3DataStore datastore) {
+        AWSCredentialsProviderChain credentials = S3Service.createAWSCredentials(datastore);
+        return AmazonS3ClientBuilder.standard().withRegion(Regions.DEFAULT_REGION).withCredentials(credentials).build();
+    }
+
+    // TODO check if region if valid
```
laurentgir: remove

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
rdubois: The `datastore` string has been externalized to a constant.

connections/connection-s3/src/main/java/org/talend/components/fileio/s3/S3Service.java
rdubois: Done

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
```diff
@@ -0,0 +1,101 @@
...
+ * This BaseParameterEnricher is responsible for adding some metadata to the fields annotated with @DatastoreRef.
+ *
+ * These metadata will be used by client applications to change the default display behavior of the annotated fields.
+ * For example, in Pipeline Designer, the annotated field will be displayed with a closed drop down list listing the
+ * connections of the right kind, available in the Pipeline Designer application.
+ */
+@Slf4j
+public class DatastoreRefEnricher extends BaseParameterEnricher {
+
+    private static final String META_PREFIX = "tcomp::configurationtyperef::";
```
rmannibucau: Hi, maybe move to META_PREFIX="connectors::extensions::xxxx" instead of reusing "tcomp::" which is the internal naming prefix

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
```diff
@@ -0,0 +1,101 @@
...
+                        + parameterName + "'" + "\nCheck that you have a valid package-info file."));
+            } else {
+                Class clazz = (Class) parameterType;
+                if (!clazz.isAnnotationPresent(DataStore.class)) {
+                    log.error("Invalid DatastoreRef.",
+                            new IllegalStateException("A @DatastoreRef can only be used on @Datastore parameters"));
+                } else {
+                    final String name = ((DataStore) clazz.getAnnotation(DataStore.class)).value();
+                    final String family = c.family();
+                    final HashMap<String, String> metas = new HashMap<>();
```
rmannibucau: guess you should rather use configtype identifier there instead of trying to recreate the key, this wouldn't be stable with extensions compared to using natural id.

connections/connection-s3/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+
+    <properties>
+        <aws-java-sdk-s3.version>1.11.497</aws-java-sdk-s3.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.4</version>
```
rmannibucau: provided but I think it is in the parent so no need actually ;)

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
rdubois: +1, it has been changed.

connections/connection-s3/pom.xml
rdubois: Good catch. It's been removed.

component-extension/src/main/java/org/talend/components/extension/DatastoreRefEnricher.java
rdubois: Changes have been applied according to your comment. There was indeed a bug with the implementation. 👍

PR #219, opened 2019-10-24
feat(TDI-42349): prevent insert operation use temp table

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-42349

Improve Insert operation, by preventing temporary table usage.


jdbc/src/main/java/org/talend/components/jdbc/output/statement/operations/snowflake/SnowflakeInsert.java
```diff
@@ -42,19 +42,9 @@ public SnowflakeInsert(Platform platform, OutputConfig configuration, I18nMessag
         final List<Reject> rejects = new ArrayList<>();
         try (final Connection connection = dataSource.getConnection()) {
             final String tableName = getConfiguration().getDataset().getTableName();
-            final String tmpTableName = tmpTableName(tableName);
```
gonchar-ivan: please, clean imports

mbasiuk-talend: removed. Thanks! I've updated my IDE settings.

PR #227, opened 2019-10-30
fix(TDI-43114): handle incoming null values in CouchbaseOuput

undx:
https://jira.talendforge.org/browse/TDI-43114 - handle null values in CouchbaseOutput
We add the same issue in azure-dls-gen2 with the JsonConverter and Record processing.

### DIFF
--- a/couchbase/src/test/java/org/talend/components/couchbase/output/CouchbaseOutputTest.java
+++ b/couchbase/src/test/java/org/talend/components/couchbase/output/CouchbaseOutputTest.java
@@ -12,14 +12,10 @@
  */
 package org.talend.components.couchbase.output;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collector;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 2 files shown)


couchbase/src/test/java/org/talend/components/couchbase/output/CouchbaseOutputTest.java
```diff
@@ -302,4 +303,22 @@ private CouchbaseOutputConfiguration getOutputConfiguration() {
...
+        byte[] bytes = "aloha".getBytes(Charset.defaultCharset());
+        String idValue = "fixBytes";
+        Record test = recordBuilderFactory.newRecordBuilder().withString("ID", idValue).withInt("id", 101)
+                .withString("name", "kamikaze").withBytes("byties", bytes).build();
+        CouchbaseOutput couch = new CouchbaseOutput(getOutputConfiguration(), null, null);
+        JsonDocument jsonDoc = couch.toJsonDocument("ID", test);
+        assertEquals(idValue, jsonDoc.id());
+        JsonObject jsonObject = jsonDoc.content();
+        assertEquals(101, jsonObject.getInt("id"));
+        assertEquals("kamikaze", jsonObject.getString("name"));
```
slushatel: maybe we do not have to check other types in 'Bytes' test?

undx:
In fact, I merged both pr's commits for testing both issues in the Studio and unfortunately pushed commits. You're right will revert a part and continue splitted prs.

couchbase/src/test/java/org/talend/components/couchbase/output/CouchbaseOutputTest.java
undx: sure, but that's harmless asserts.

PR #228, opened 2019-10-30
fix(TDI-43115): handle BYTES type in CouchbaseOuput

undx:
https://jira.talendforge.org/browse/TDI-43115 - handle BYTES type in CouchbaseOuput
BYTES type is not handed and throws an IllegalArgumentException. Add a default implementation that works in studio.

### DIFF
--- a/couchbase/src/main/java/org/talend/components/couchbase/output/CouchbaseOutput.java
+++ b/couchbase/src/main/java/org/talend/components/couchbase/output/CouchbaseOutput.java
@@ -17,7 +17,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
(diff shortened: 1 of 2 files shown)


couchbase/src/main/java/org/talend/components/couchbase/output/CouchbaseOutput.java
```diff
@@ -139,7 +143,10 @@ private Object jsonValueFromRecordValue(Schema.Entry entry, Record record) {
         case LONG:
             return record.getLong(entryName);
         case BYTES:
-            throw new IllegalArgumentException("BYTES is unsupported");
+            byte[] bytes = record.getBytes(entryName);
+            Integer[] intArray = IntStream.range(0, bytes.length).map(i -> bytes[i]).boxed().collect(Collectors.toList())
+                    .toArray(new Integer[bytes.length]);
+            return JsonArray.from(intArray);
```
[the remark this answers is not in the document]
undx: I've to check with tck 1.1.14 in the studio.
The guess schema autodetects a string type for now.
This is, for now a proposal on writing byte[] with CouchbaseOutput, decision on which  final type written is done yet by Pierre. But Pierre does not want a String type (prone to errors w/ encodings).

couchbase/src/main/java/org/talend/components/couchbase/output/CouchbaseOutput.java
```diff
@@ -139,7 +142,7 @@ private Object jsonValueFromRecordValue(Schema.Entry entry, Record record) {
         case LONG:
             return record.getLong(entryName);
         case BYTES:
-            throw new IllegalArgumentException("BYTES is unsupported");
+            return com.couchbase.client.core.utils.Base64.encode(record.getBytes(entryName));
```
slushatel: Can we use Base64 from java.util?
import java.util.Base64;
byte[] encodedBytes = Base64.getEncoder().encode("Test".getBytes());

undx: We can but I had previously some curious issues when switching java version (8>11) and using base64... I hope with using couchbase sdk we have a constant behavior.

slushatel: strange)

PR #231, opened 2019-11-04
chore(TPS-3537): remove snapshot deps

undx:
set commons-csv from snapshot to release for release process


clesaec:
reste un 1.2.3-SNAPSHOT sur le pom.xml, je sais pas s'il va aimer ....

undx:
jenkins release process should do the trick ?

PR #242, opened 2019-11-25
fix(TDI-43214): improved Release stage - backport maintenance/1.3

undx:


clesaec:
Change image use (l 43) with image: 'artifactory.datapwn.com/tlnd-docker-dev/talend/common/tsbi/jdk8-builder-base:1.12.0-20191022071355'
(like [here](https://github.com/Talend/devtools/blob/master/nexus/Jenkinsfile#L31))

PR #244, opened 2019-11-25
Ypiel/tdi 41956 rest tacokit connector

ypiel-talend:
Tacokit REST connector first lot
https://jira.talendforge.org/browse/TDI-42996

Reviewed on PR to master https://github.com/Talend/connectors-se/pull/224


ypiel-talend:
PR validated within the PR on master. I merge first on maintenance/1.3 since master is already using tacokit 1.1.15 which is not compatible with the connector REST (little code fix to do to be compatible with 1.1.15 about redirections).

PR #254, opened 2019-12-02
fix(TDI-42987): [Tacokit][Azure blob storage] doesn't work for core.chinacloudapi.cn

wwang-talend:


azure/azureblob/src/main/java/org/talend/components/azure/runtime/input/ParquetBlobFileReader.java
```diff
@@ -84,15 +81,17 @@ private void initConfig() {
             hadoopConfig = new Configuration();
             hadoopConfig.set(AZURE_FILESYSTEM_PROPERTY_KEY, AZURE_FILESYSTEM_PROPERTY_VALUE);
             if (getConfig().getConnection().isUseAzureSharedSignature()) {
-                Matcher mather = sasPattern
-                        .matcher(getConfig().getConnection().getSignatureConnection().getAzureSharedAccessSignature());
-                accountName = mather.group(2);
```
wwang-talend: seems this code never been tested

wwang-talend:
the test failed above is caused by other thing, not current fix

PR #257, opened 2019-12-03
fix(TDI-43299): marketo relative date

undx:
https://jira.talendforge.org/browse/TDI-43299


marketo/src/main/resources/org/talend/components/marketo/dataset/Messages.properties
```diff
@@ -32,3 +32,10 @@ OutputAction.createOrUpdate._displayName=Create or Update
 OutputAction.delete._displayName=Delete
 OutputAction.sync._displayName=Synchronize
 OutputAction.updateOnly._displayName=Update only
+DateTimeRelative.PERIOD_AGO_1M._displayName=last month
```
ypiel-talend: Should we put 1st letter uppercase ?

undx: It has to read as _Modified within the ... last month_
There were a wording session about that.

ypiel-talend: Ok so those labels are parameters of some others, ok.

PR #259, opened 2019-12-04
fix(TDI-43303): remove default value for sinceDateTimeAbsolute

undx:
https://jira.talendforge.org/browse/TDI-43303


marketo/src/main/java/org/talend/components/marketo/dataset/MarketoDataSet.java
```diff
@@ -89,8 +87,7 @@
             @ActiveIf(target = "dateTimeMode", value = { "absolute" }) })
     @Validable(VALIDATION_DATETIME_PATTERN)
     @Documentation("Since Absolute Date Time")
```
ypiel-talend: Don't upper case each word : "Since absolute date time"

marketo/src/main/java/org/talend/components/marketo/dataset/MarketoDataSet.java
```diff
@@ -89,8 +87,7 @@
...
-    private String sinceDateTimeAbsolute = ZonedDateTime.now().minusMonths(2)
-            .format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"));
+    private String sinceDateTimeAbsolute;
```
ypiel-talend: Should it become @Required ?

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
```diff
@@ -308,7 +308,7 @@ public ValidationResult validateList(final List<String> list) {
 
     @AsyncValidation(VALIDATION_DATETIME_PATTERN)
     public ValidationResult validateDateTimePattern(final String date) {
```
ypiel-talend: Should we check that the content is a well formatted datetime pattern as expected ?

marketo/src/main/java/org/talend/components/marketo/dataset/MarketoDataSet.java
undx: It cannot yet because it doesn't work for dataset (@required and @activeIf used together).

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
undx: Yes, because the format should be respected on Marketo side.

marketo/src/main/java/org/talend/components/marketo/dataset/MarketoDataSet.java
undx: will downcase.

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
ypiel-talend: yes but you don't validate the content only if null or formatted no ?

marketo/src/main/java/org/talend/components/marketo/dataset/MarketoDataSet.java
ypiel-talend: yes you're right https://jira.talendforge.org/browse/TDC-3205

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
ypiel-talend: Sorry github folded the method, you well check the date format

undx: Check code below, we validate that input is well matching a datetime as Marketo expects.

PR #265, opened 2019-12-11
fix(TDI-43335):change error message

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-43335


pyzhou-talend:
This will not fix the root problem. Will fix in another ticket TDI-43376

PR #284, opened 2019-12-27
Pyzhou/tdi 43097 snowflake remove shutdown hook

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-43097


pyzhou-talend:
Thanks a lot for helping with this issue. I will remove the static util mentioned. About the rollback, yes, good leaving it for more discussion and open up a new Jira for it.

PR #286, opened 2020-01-02
TDI-43458 : maven decrypter to access workday credentials

clesaec:
Usage of MavenDecrypter when passthrough option to true on pom.xml


workday/src/test/java/org/talend/components/workday/service/ConfigHelper.java
```diff
@@ -29,6 +35,21 @@ public static Properties workdayProps() {
         try (InputStream in = Thread.currentThread().getContextClassLoader().getResourceAsStream("workdayConfig.properties")) {
             Properties wkprops = new Properties();
             wkprops.load(in);
+
+        /*    final String property = System.getProperty("talend.junit.http.passthrough"); // property defined in pom.xml.
+            if ("true".equals(property)) {
+                MavenDecrypter decrypter = new MavenDecrypter();
```
undx: It's better to use the following best practice 
```
    @Rule
    public final MavenDecrypterRule mavenDecrypterRule = new MavenDecrypterRule(this);

    @DecryptedServer("workday")
    private Server server;
```
Then, use `server.getUsername()` and `server.getPassword()`.

workday/src/test/resources/talend/testing/http/org.talend.components.workday.input.RAASProducerTest_nextError.json
```diff
@@ -19,7 +19,7 @@
         "X-Content-Type-Options":"nosniff",
         "Content-Security-Policy":"default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data:",
         "Connection":"keep-alive",
-        "Set-Cookie":"AWSALB=kMoxPioYPQpL4yu6NI1vNzrI+PnCK8Zyes/fKgBUCUMXWcDISpJr596HEypw; Expires=Fri, 27 Dec 2019 17:07:21 GMT; Path=/",
+        "Set-Cookie":"AWSALB=kMoxPioYPQpL4yu6NI1vNzrI+PnCK8Zyes/fKgBUCUMXWcDISpJr596HEypw; Expires=Fri, 27 Dec 2039 17:07:21 GMT; Path=/",
```
undx: Question: what's this cookie? Auth cookie or cookite to get a token? If so, can we use it for logging to WD API ? (expire date is 2039)

workday/src/test/java/org/talend/components/workday/service/ConfigHelper.java
undx: Not sure that's needed but may need also : 
```
private final MavenDecrypter mavenDecrypter = new MavenDecrypter();
``` 
(after reading the doc)

clesaec: @Rule is a JUnit4 feature i think ?

undx: Hum, yeah you're absolutely right !
Sorry, here's a j5 example: 

```
@WithMavenServers
@WithComponents("org.talend.components.marketo")
public class MarketoBaseTest {

    @ClassRule
    public static final SimpleComponentRule component = new SimpleComponentRule("org.talend.components.marketo");

    @DecryptedServer(value = "marketo-nocrm")
    protected Server serverNoCrm;

    @DecryptedServer(value = "marketo-nocrm-instance")
    protected Server serverNoCrmInstance;
```

clesaec: @ClassRule is also JUnit4 feature.
I will try with `@WithMavenServers` and `@DecryptedServer` annotations (it seems very similar).

undx: not needed in this example, was just a copy/paste...
`@WithMavenServers` and `@DecryptedServer`

clesaec: Done

workday/src/test/resources/talend/testing/http/org.talend.components.workday.input.RAASProducerTest_nextError.json
clesaec: This info is not usefull, i removed it.

workday/src/test/java/org/talend/components/workday/WorkdayBaseTest.java
```diff
@@ -0,0 +1,45 @@
...
+        } catch (IOException ex) {
+            throw new UncheckedIOException(ex);
+        }
+    }
+
+    protected WorkdayDataStore buildDataStore() {
+        Properties props = this.workdayProps();
+        WorkdayDataStore wds = new WorkdayDataStore();
+        wds.setClientId(serverWorkday.getUsername());
+        wds.setClientSecret(props.getProperty(serverWorkday.getPassword()));
```
undx: I don't understand where is the real secret? Is it in the properties file with key of value of the maven settings file ? Keep the secret key in mvn, no ? 
In your settings.xml, you can have multiple servers (so multiple @DecryptedServer), you can use that feature to get rid of your properties file and remove it from VCS.

clesaec: I can remove clientId and clientSecret from properties files (it's now unused).
But, i have top keep tenant with it's value.
Server class get by maven contains only 2 attributes for each server, and i need at least 3 (clientId, secret, tenant alias). Secret is mostly on client secret, so put this on settings.xml.

clesaec: (Yesd, i just see pb :))

clesaec: correction done

undx: I know, that's why you may use more than one Server.
For instance, in Marketo we have 2 servers in settings.xml :
one with the specific endpoint and another one for creds.

clesaec: Can we consider to add JsonObject (or Properties) to Server class to add any others parameters. (Here, we would have 3 Server instances to describe one server access). 
In the mean time, is it really better to have several Server in setting.xml (may be hard to understand ?)

undx: I think that you should be able to have `CDATA` section in one of the two fields on the server entry as it's a xml file...

undx: Discussed on slack, maybe plan to extend Server class to handle other Server's schema elements.

clesaec: done with 2 server for the moment ...

undx: ok, we'll later how to handle this more cleanly this kind of case.

PR #287, opened 2020-01-02
TDI-43483 : substitutor with regexp solution

clesaec:
substitutor with regexp solutions and adaptation for REST module (to avoid systematic regexp compilation that have a great time cost)


common/src/main/java/org/talend/components/common/text/Substitutor.java
```diff
@@ -12,197 +12,214 @@
...
+ * 
+ * it doesn't treat recursive replacement: '${greeting}' with a function 
+ *     'greeting' -> '${hello} ${world:-tdi}" and 'hello' -> 'hi' will render '${hello} ${world:-tdi}'.
  */
 public class Substitutor {
 
-    private final char escape = '\\';
+    private static final char ESCAPE = '\\';
 
-    private final char[] delimiter = ":-".toCharArray();
```
rlecomte-talend: field name should be in uppercase, it's a constant

clesaec: renamed

PR #289, opened 2020-01-03
feat(TDI-42337) : Add svg icons in connectors resources

ypiel-talend:
https://jira.talendforge.org/browse/TDI-42337
Icons has to be retrieved by calling component server. Icons must be in resources/icons.
Png icons for studio are generated from svg ones if the png icons doesn't exist.


Voievidko:
in and out icons are same. Is it ok?

ypiel-talend:
For TPD we can keep same icon for in/out. We have to keep different icons for studio.

PR #290, opened 2020-01-06
fix(TDI-43497): handle more lists than batch limit

undx:
https://jira.talendforge.org/browse/TDI-43497

### DIFF
--- a/marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
+++ b/marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
@@ -72,7 +72,7 @@ public class LeadSource extends MarketoSource {
     private JsonObject getLeadsByListId() {
         Integer listId = Integer.parseInt(configuration.getDataSet().getListId());
         String fields = schema.keySet().stream().collect(Collectors.joining(","));
-        return handleResponse(listClient.getLeadsByListId(accessToken, nextPageToken, listId, fields));
+        return marketoService.handleResponse(listClient.getLeadsByListId(accessToken, nextPageToken, listId, fields));
     }
 
     private Boolean isLeadUrlSizeGreaterThan8k(String filterType, String filterValues, String fields) {
@@ -132,15 +132,16 @@ public class LeadSource extends MarketoSource {
             activityTypeIds = configuration.getDataSet().getActivityTypeIds().stream().collect(joining(","));
         }
         String listId = configuration.getDataSet().getListId();
-        return handleResponse(leadClient.getLeadActivities(accessToken, nextPageToken, activityTypeIds, "", listId, ""));
+        return marketoService
+                .handleResponse(leadClient.getLeadActivities(accessToken, nextPageToken, activityTypeIds, "", listId, ""));
     }
 
     public JsonObject getActivities() {
-        return handleResponse(leadClient.getActivities(accessToken));
+        return marketoService.handleResponse(leadClient.getActivities(accessToken));
     }
 
     public String getPagingToken(String dateTime) {
-        return handleResponse(leadClient.getPagingToken(accessToken, dateTime)).getString(ATTR_NEXT_PAGE_TOKEN);
+        return marketoService.handleResponse(leadClient.getPagingToken(accessToken, dateTime)).getString(ATTR_NEXT_PAGE_TOKEN);
     }
 
 }
--- a/marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
+++ b/marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
@@ -72,7 +72,7 @@ public class LeadSource extends MarketoSource {
     private JsonObject getLeadsByListId() {
         Integer listId = Integer.parseInt(configuration.getDataSet().getListId());
         String fields = schema.keySet().stream().collect(Collectors.joining(","));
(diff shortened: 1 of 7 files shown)


undx:
and a bit more shorter !!! Java naming syndrom :-)

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
```diff
@@ -203,15 +207,25 @@ public SuggestionValues getActivities(@Option final MarketoDataStore dataStore)
     @Suggestions(LIST_NAMES)
     public SuggestionValues getListNames(@Option final MarketoDataStore dataStore) {
         log.debug("[getListNames] {}.", dataStore);
+        List<Item> lists = new ArrayList<>();
+        String nextPageToken = null;
+        Predicate<JsonObject> hasResults = j -> j.getJsonArray(ATTR_RESULT) != null;
+        Predicate<JsonObject> hasNextPageToken = j -> j.getString(ATTR_NEXT_PAGE_TOKEN, null) != null;
+        Consumer<JsonObject> listConsumer = l -> lists.add(new Item(String.valueOf(l.getInt(ATTR_ID)), l.getString(ATTR_NAME)));
```
clesaec: JsonObject.getInt(ATTR_ID) throw an Exception if not exists.
May be better to write a private function that transform JsonObject to Item ?

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
```diff
@@ -203,15 +207,25 @@ public SuggestionValues getActivities(@Option final MarketoDataStore dataStore)
...
-                listNames.add(new SuggestionValues.Item(String.valueOf(l.getInt(ATTR_ID)), l.getString(ATTR_NAME)));
+            JsonObject result = handleResponse(listClient.getLists(aToken, nextPageToken, null, "", "", ""));
+            nextPageToken = result.getString(ATTR_NEXT_PAGE_TOKEN, null);
+            if (hasResults.test(result)) {
+                result.getJsonArray(ATTR_RESULT).getValuesAs(JsonObject.class).forEach(listConsumer);
+            }
+            while (hasNextPageToken.test(result)) {
+                result = handleResponse(listClient.getLists(aToken, nextPageToken, null, "", "", ""));
+                nextPageToken = result.getString(ATTR_NEXT_PAGE_TOKEN, null);
+                result.getJsonArray(ATTR_RESULT).getValuesAs(JsonObject.class).forEach(listConsumer);
```
clesaec: Miss "hasResults.test(result)"
think about create a private function to encaspulate repeat code

```            
      JsonObject result = handleResponse(listClient.getLists(aToken, nextPageToken, null, "", "", ""));
      nextPageToken = result.getString(ATTR_NEXT_PAGE_TOKEN, null);
      if (hasResults.test(result)) {
            return result.getJsonArray(ATTR_RESULT).getValuesAs(JsonObject.class);
      }
      return Collections.emptyList();
```
call it `getNextPage`

undx: if nextpagetoken is defined and there's not error in request (handleResponse), we don't need this predicate again. It's only needed for the first call which may be successfull but have no entries.

undx: Wanted to preserve the local scope of variables. we'll look to refactor anyway.

marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
```diff
@@ -132,15 +132,16 @@ private JsonObject getLeadActivities() {
             activityTypeIds = configuration.getDataSet().getActivityTypeIds().stream().collect(joining(","));
         }
         String listId = configuration.getDataSet().getListId();
-        return handleResponse(leadClient.getLeadActivities(accessToken, nextPageToken, activityTypeIds, "", listId, ""));
+        return marketoService
+                .handleResponse(leadClient.getLeadActivities(accessToken, nextPageToken, activityTypeIds, "", listId, ""));
```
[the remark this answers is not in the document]
undx: Will have a look but pagination is already handled in source comp.

marketo/src/main/java/org/talend/components/marketo/input/LeadSource.java
```diff
@@ -72,7 +72,7 @@ public JsonObject runAction() {
     private JsonObject getLeadsByListId() {
         Integer listId = Integer.parseInt(configuration.getDataSet().getListId());
         String fields = schema.keySet().stream().collect(Collectors.joining(","));
-        return handleResponse(listClient.getLeadsByListId(accessToken, nextPageToken, listId, fields));
+        return marketoService.handleResponse(listClient.getLeadsByListId(accessToken, nextPageToken, listId, fields));
```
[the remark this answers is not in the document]
undx: The scope of these generic handling functions have broaden to outside the scope of SourceOrSink.

marketo/src/main/java/org/talend/components/marketo/service/UIActionService.java
undx: Actually, if the previous checks were successful, there's no reason that it do not exists. Mkto maintains integrity of its lists.

PR #291, opened 2020-01-07
TDI-43491 : fix NPE on datetime

rlecomte-talend:


bigquery/src/main/java/org/talend/components/bigquery/output/TacoKitRecordToTableRowConverter.java
```diff
@@ -132,7 +132,9 @@ public TacoKitRecordToTableRowConverter(com.google.cloud.bigquery.Schema tableSc
                     tableRow.put(fieldName, input.getBoolean(fieldName));
                     break;
                 case DATETIME:
-                    tableRow.put(fieldName, getDateFunction(field.getType()).apply(input.getDateTime(fieldName)));
+                    tableRow.put(fieldName, Optional.ofNullable(input.getDateTime(fieldName))
+                            .map( dt -> getDateFunction(field.getType()).apply(dt))
+                            .orElse(null));
```
undx: Do you think that we may have some nulls in array (L107) ?

bigquery/src/test/java/org/talend/components/bigquery/output/TacoKitRecordToTableRowConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.talend.sdk.component.runtime.record.RecordBuilderFactoryImpl;
+
+import java.util.Date;
+import java.util.Locale;
+import java.util.Map;
+
+public class TacoKitRecordToTableRowConverterTest {
+
+    @Test
+    public void testTDI_43491() {
```
undx: rename method to match test case.

bigquery/src/main/java/org/talend/components/bigquery/output/TacoKitRecordToTableRowConverter.java
rlecomte-talend: I'll double check this one too

bigquery/src/main/java/org/talend/components/bigquery/output/TacoKitRecordToTableRowConverter.java
```diff
@@ -24,10 +24,7 @@
 
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
undx: latest comment ;-) 
avoid star import

rlecomte-talend: Have to go deep inside Intellij settings to correct that

undx: sent idea config settings through slack

PR #293, opened 2020-01-08
fix(TDI-43525): ClassCastException on short type JDBC

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-43525


jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
```diff
@@ -128,7 +128,7 @@ private void addField(final Schema.Builder builder, final ResultSetMetaData meta
             case java.sql.Types.SMALLINT:
             case java.sql.Types.TINYINT:
             case java.sql.Types.INTEGER:
-                if (javaType.equals(Integer.class.getName())) {
+                if (javaType.equals(Integer.class.getName()) || Short.class.getName().equals(javaType)) {
```
wwang-talend: not friend to 强迫症, but ok

PR #302, opened 2020-01-13
chore(TDI-43439): add source clear scan

pteyssier:
- also bump tsbi image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ spec:
     containers:
         -
             name: main
-            image: 'artifactory.datapwn.com/tlnd-docker-dev/talend/common/tsbi/jdk8-builder-base:1.16.0-20200106163950'
+            image: 'artifactory.datapwn.com/tlnd-docker-dev/talend/common/tsbi/jdk8-svc-springboot-builder:1.14.0-2.1-20191203093421'
             command: [cat]
             tty: true
             volumeMounts: [{name: docker, mountPath: /var/run/docker.sock}, {name: m2main, mountPath: /root/.m2/repository}]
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -71,11 +71,12 @@ spec:
 
     triggers {
         cron(env.BRANCH_NAME == "master" ? "@daily" : "")
+        parameterizedCron(env.BRANCH_NAME == "master" ? "0 9 * * 0 %SCA_SCAN=true" : "") // Sunday at 9am. Only on master
     }
 
     parameters {
         choice(name: 'Action', 
-               choices: [ 'STANDARD', 'PUSH_TO_XTM', 'DEPLOY_FROM_XTM', 'RELEASE' ],
+               choices: [ 'STANDARD', 'PUSH_TO_XTM', 'DEPLOY_FROM_XTM', 'SCAN', 'RELEASE' ],
```
ypiel-talend: Victor told me we can have several Jenkinsfile. So maybe we can split in several jobs instead of having choices ?

Jenkinsfile
```diff
@@ -207,6 +208,20 @@ spec:
...
             }
         }
+        stage("SourceClear analysis") {
+            when {
+                expression { params.Action == 'SCAN' || ${SCA_SCAN} == true }
+            }
+            steps {
+                container('main') {
+                    withCredentials([string(credentialsId: 'veracode-token', variable: 'SRCCLR_API_TOKEN')]) {
+                        sh '''#!/bin/bash
```
ypiel-talend: I don't think the given interpreter within sheebang is really used here.

pteyssier: Maybe. but that's the line used by other team and tested by arch team.

Jenkinsfile
pteyssier: It's a valid request (and as old as the xtm are there), but not related to this PR.

PR #318, opened 2020-02-06
Ypiel/tdi 43542 rest body to records

ypiel-talend:
https://jira.talendforge.org/browse/TDI-43542

Body is no more return as String bu as jsonObject or several rows to be computed by other application like dataprep.


rest/src/main/java/org/talend/components/rest/service/RestService.java
```diff
@@ -218,39 +220,34 @@ public String setPathParams(String resource, boolean hasPathParams, Map<String,
...
-                    .headers().entrySet().stream().map(e -> recordBuilderFactory.newRecordBuilder(headerElementSchema)
-                            .withString("key", e.getKey()).withString("value", String.join(",", e.getValue())).build())
-                    .collect(Collectors.toList());
-
-            Schema.Entry.Builder arrayEntryBuilder = recordBuilderFactory.newEntryBuilder();
-            builder.withArray(arrayEntryBuilder.withName("headers").withType(Schema.Type.ARRAY).withNullable(false)
-                    .withElementSchema(headerElementSchema).build(), headers);
+
+        Map<String, List<String>> responseHeaders = resp.headers();
+        Map<String, String> headers = new HashMap<>();
```
mbasiuk-talend: Can be, or may not be updated with such suggestion:
```
ofNullable(resp.headers())
                .orElseGet(Collections::emptyMap)
                .entrySet()
                .stream()
                .collect(toMap(entry -> entry.getKey(), entry -> join(",", entry.getValue())));
```

rest/src/main/java/org/talend/components/rest/service/RestService.java
```diff
@@ -218,39 +220,34 @@ public String setPathParams(String resource, boolean hasPathParams, Map<String,
...
+
+    private static String getBody(final Response<byte[]> resp) {
         String encoding = ContentType.getCharsetName(resp);
         String receivedBody = (encoding == null) ? //
                 new String(Optional.ofNullable(resp.body()).orElse(new byte[0])) : //
                 new String(Optional.ofNullable(resp.body()).orElse(new byte[0]), Charset.forName(encoding));
-        builder.withString("body", receivedBody);
-
-        return builder.build();
+        return receivedBody;
```
mbasiuk-talend: Maybe extract local variable `byte[] bytes = Optional.ofNullable(resp.body()).orElse(new byte[0])`

rest/src/main/java/org/talend/components/rest/virtual/ComplexRestEmitter.java
```diff
@@ -0,0 +1,84 @@
...
+    public Object next() {
+        if (items == null) {
+            items = new LinkedList<>();
+            final RestEmitter delegateSource = new RestEmitter(configuration.getRestConfiguration(), client);
+            final CompletePayload global = delegateSource.next();
+
+            final boolean isJson = !String.class.isInstance(global.getBody());
+            final boolean isCompletePayload = configuration.getRestConfiguration().getDataset().isCompletePayload();
+
+            if (isJson) {
```
mbasiuk-talend: maybe split this method with extra methods like: processJsonResponse

rest/src/main/java/org/talend/components/rest/configuration/Dataset.java
```diff
@@ -108,6 +107,10 @@
     @Documentation("Request body")
     private RequestBody body;
 
+    @Option
+    @Documentation("Return complete payload as a record")
```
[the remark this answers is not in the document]
ypiel-talend: Better to have boolean and not Boolean to avoid NPE in taht case.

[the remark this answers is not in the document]
ypiel-talend: Change Boolean to boolean in others

rest/src/main/java/org/talend/components/rest/processor/JSonExtractor.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2006-2019 Talend Inc. - www.talend.com
```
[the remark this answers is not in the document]
ypiel-talend: Will do a bulk change later on all source files...

rest/src/main/java/org/talend/components/rest/processor/JSonExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.json.JsonObject;
+import javax.json.JsonPointer;
+import javax.json.JsonValue;
+import java.util.Optional;
+
+/**
+ * Processor
+ */
+/*
+ * @Version
```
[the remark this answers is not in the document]
ypiel-talend: It was to keep logic for future to have 'virtual' connectors.

rest/src/main/java/org/talend/components/rest/service/RestService.java
ypiel-talend: Yes thanks

rest/src/main/java/org/talend/components/rest/service/RestService.java
ypiel-talend: Done

PR #324, opened 2020-02-13
fix(TDI43698) - locales version ranges

clesaec:
Fix locales versions - [TDI-43698](https://jira.talendforge.org/browse/TDI-43698)


gonchar-ivan:
Looks ok for me if we are sure that locales 1.6.6+ won't be release during Studio release

PR #329, opened 2020-02-18
Clesaec/tdi 43483 rest substitutor

clesaec:
Substitutor refactor (impact on REST component)


common/src/test/java/org/talend/components/common/text/SubstitutorTest.java
```diff
@@ -36,18 +39,29 @@ void testSubstitutor(final String prefix, final String suffix, final String valu
...
-        Assert.assertEquals(expected, transformed);
+        Assertions.assertEquals(expected, transformed);
 
     }
 
+    @Test
+    public void testFinder() {
+        Substitutor.KeyFinder finder = new Substitutor.KeyFinder("${", "}");
+        Iterator<Substitutor.FindResult> res = finder.search("${hello} ${world:-tdi}");
+        Assertions.assertTrue(res.hasNext());
```
ypiel-talend: Why several hasNext ?

clesaec: Test re-entrant method property.

ypiel-talend: I rememeber you already answered me, it was to be sure that the iterator impl don't consume the next within the hasNext.

PR #335, opened 2020-02-20
fix(TDI-43720): Parse headers in excel html

dmytro-sylaiev:


azure/azureblob/src/main/java/org/talend/components/azure/runtime/input/ExcelHTMLBlobFileReader.java
```diff
@@ -63,6 +63,10 @@ protected void readItem() {
                 Element body = document.body();
                 Elements rows = body.getElementsByTag("tr");
                 rowIterator = rows.iterator();
+                if (rows.first().getElementsByTag("th").size() > 0) {
```
chmyga: you can use HTMLConverter isHeaderRecord method or move that method to some util class to make the code a bit prettier.

PR #337, opened 2020-02-24
Follow redirect

clesaec:


workday/src/main/java/org/talend/components/workday/service/AccessTokenConfigurer.java
```diff
@@ -24,6 +24,12 @@
     public void configure(Connection connection, ConfigurerConfiguration configuration) {
         log.debug("[configure] [{}] {}", connection.getMethod(), connection.getUrl());
         connection //
-                .withHeader("Content-Type", CONTENT_TYPE_APPLICATION_X_WWW_FORM_URLENCODED).withoutFollowRedirects();
+                .withHeader("Content-Type", CONTENT_TYPE_APPLICATION_X_WWW_FORM_URLENCODED);
+
+        try {
+            connection.withoutFollowRedirects();
+        } catch (NoSuchMethodError ex) {
+            log.warn("can't do withoutFollowRedirects");
```
undx: `withoutFollowRedirects operation unsupported.` ?

clesaec: Yes, when tck version for compilation is not the same as tck version in remote engine.

undx: I was talking about the log.warn message content ;-)

clesaec: different than 'can't do withoutFollowRedirects' ?

undx: a little bit nicer :-)

clesaec: Changed even if it's same meaning.

PR #338, opened 2020-02-25
fix(TDI-43519): Add chunk size parameter on snowflake component

wwang-talend:


wwang-talend:
test failed as can't reach the maven repositories, not caused by current fix, and have passed local mvn clean install

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
```diff
@@ -120,6 +119,12 @@
     @Documentation("Rewrite batched statements, to execute one statement per batch combining values in the sql query")
     private boolean rewriteBatchedStatements = true;
 
+    @Option
+    @ActiveIf(target = "../dataset.connection.dbType", value = "Snowflake")
+    @Documentation("set the max temp file size(byte) for snowflake put and copy command, if meet it, will consume it at once and create new")
+    @Min(0)
+    private long chunkSize = 50 * 1024 * 1024;
```
pyzhou-talend: maybe better to let user fill in with MB rather than byte

jdbc/src/main/resources/org/talend/components/jdbc/configuration/Messages.properties
```diff
@@ -17,6 +17,7 @@ OutputConfig.distributionStrategy._displayName=Distribution strategy
 OutputConfig.distributionStrategy._placeholder=
 OutputConfig.distributionKeys._displayName=Distribution keys
 OutputConfig.distributionKeys._placeholder=
+OutputConfig.chunkSize._displayName=Chunk size (bytes)
```
pyzhou-talend: please add the placeholder

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
wwang-talend: this follow the jira desc, keep the same with studio one

jdbc/src/main/resources/org/talend/components/jdbc/configuration/Messages.properties
wwang-talend: done

PR #341, opened 2020-02-26
Rlecomte/tdi 43745 pubsub text

rlecomte-talend:


pubsub/src/main/resources/org/talend/components/pubsub/dataset/Messages.properties
```diff
@@ -11,6 +11,8 @@ PubSubDataSet.subscription._displayName = Subscription
 PubSubDataSet.subscription._placeholder = Id of subscription
 PubSubDataSet.avroSchema._displayName = Avro schema
 PubSubDataSet.avroSchema._placeholder = Expected schema
+PubSubDataSet.pathToText._displayName = Path to text
+PubSubDataSet.pathToText._placeholder = Path to text in record
```
clesaec: Place holder should be an example; no ? "/field1/field2" ?

rlecomte-talend: It could, but there's already a default value to show the format (in fact, the placeholder will appear only if user remove the default value).

pubsub/src/main/java/org/talend/components/pubsub/dataset/PubSubDataSet.java
```diff
@@ -79,6 +81,13 @@
     @Documentation("Avro schema")
     private String avroSchema;
 
+    @Option
+    @DefaultValue("/content")
+    @ActiveIf(target = "valueFormat", value = { "TEXT" })
```
undx: Missed something? This is about text or json as valueformat is handling both? 
Text using jsonPointer, maybe you should reword the format and related fields....

rlecomte-talend: This PR is about plain text format. JsonPointer format is used to create a recordPointer to point to the incoming record field to extract text value from.

pubsub/src/main/resources/org/talend/components/pubsub/dataset/Messages.properties
undx: Agree with Christope, I'll tend to say: remove the default value and put an explicit placeholder.

rlecomte-talend: done

PR #359, opened 2020-03-11
TDI-43660 : pubsub connection - pattern for project name

rlecomte-talend:
White spaces forbidden in project name

### DIFF
--- a/pubsub/src/main/java/org/talend/components/pubsub/datastore/PubSubDataStore.java
+++ b/pubsub/src/main/java/org/talend/components/pubsub/datastore/PubSubDataStore.java
@@ -40,7 +40,7 @@ public class PubSubDataStore implements Serializable {
     @Option
     @Required
     @Documentation("Google Cloud Platform Project")
-    @Pattern("/^\\S+$/")
+    @Pattern("/^[a-z][a-z0-9\\-]*$/")
     private String projectName;
 
     @Option
--- a/pubsub/src/main/java/org/talend/components/pubsub/datastore/PubSubDataStore.java
+++ b/pubsub/src/main/java/org/talend/components/pubsub/datastore/PubSubDataStore.java
@@ -17,6 +17,7 @@ import org.talend.components.pubsub.service.PubSubService;
 import org.talend.sdk.component.api.component.Icon;
 import org.talend.sdk.component.api.configuration.Option;
 import org.talend.sdk.component.api.configuration.action.Checkable;
+import org.talend.sdk.component.api.configuration.constraint.Pattern;
 import org.talend.sdk.component.api.configuration.constraint.Required;
 import org.talend.sdk.component.api.configuration.type.DataStore;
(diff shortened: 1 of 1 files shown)


pubsub/src/main/java/org/talend/components/pubsub/datastore/PubSubDataStore.java
```diff
@@ -39,6 +40,7 @@
     @Option
     @Required
     @Documentation("Google Cloud Platform Project")
+    @Pattern("/^\\S+$/")
```
undx: Everything except space chars, very flexible. No constraints on project's names? (for instance, I know that azure has some...)

rlecomte-talend: In fact, I don't even know if whitespaces are forbidden. Documentation is not very explicit : https://cloud.google.com/resource-manager/docs/creating-managing-projects#before_you_begin

undx: So, maybe just keep the required annotation.

rlecomte-talend: Another source : https://edu.google.com/openonline/course-builder/docs/1.11/set-up-course-builder/create-a-cloud-project.html
Explicitly says no whitespace, and thta's the only rule given.

undx: OK, I read : 
> Project Name
> To be valid, a project name must
> be between 4 and 30 characters
> consist of letters, numbers, quotes, hyphens, spaces or exclamation point

Then 
> Project ID
> The project ID is a single word (no spaces) that uniquely identifies your application

So, we're talking about the project ID and not the project name. 
The label / fieldname is confusing therefore.
Maybe you should rename the option.

Also, are those kind of chars supported `#$%^&%*()`...  ?

rlecomte-talend: Already talked about it durgin development. The Google client API uses "project name" to talk about this ID... It's quite of a mess. We chose to keep it like that for now.

I also read
> The default project ID is composed of the project name and a randomly assigned number. You can customize the project ID only during project creation.
So, if project name can contains space, so can project id...

rlecomte-talend: @fdesiles made some more tests by creating new project. The rules are not the same on the project creation page... He's adding a screenshot

rlecomte-talend: "Project ID can have lowercase letters, digits or hyphens and must start with a lowercase letter". 

I'll change the regex to match this.

PR #362, opened 2020-03-16
if the key in json contains "$", can't retrieve value successfully

wwang-talend:


wwang-talend:
https://jira.talendforge.org/browse/TDI-43857

PR #366, opened 2020-03-19
feat(TDI-43868): Netsuite. To use parallelization only with Token aut…

sbovsunovskyi:
https://jira.talendforge.org/browse/TDI-43868

To use parallelization only with Token-based authentication.
All tests are enabled

### DIFF
--- a/netsuite/pom.xml
+++ b/netsuite/pom.xml
@@ -122,13 +122,6 @@
                             </classpathDependencyExcludes>
                         </configuration>
                     </execution>
-
-                    <execution>
-                        <id>surefire-beam-test</id>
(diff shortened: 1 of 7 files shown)


netsuite/pom.xml
```diff
@@ -122,13 +122,6 @@
                             </classpathDependencyExcludes>
                         </configuration>
                     </execution>
-
-                    <execution>
-                        <id>surefire-beam-test</id>
-                        <goals>
-                            <goal>test</goal>
-                        </goals>
-                    </execution>
```
gonchar-ivan: could you please explain what was purpose of this configuration?

PR #368, opened 2020-03-23
Pyzhou/tdi 43717 create cosmos db sql api

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-43717


cosmosDB/src/main/java/org/talend/components/cosmosDB/input/CosmosDBInput.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private I18nMessage i18n;
+
+    private final CosmosDBInputConfiguration configuration;
+
+    private final RecordBuilderFactory builderFactory;
+
+    private CosmosDBService service;
+
+    private DocumentClient client;
```
wwang-talend: no need to serialize, so need "transient" key. maybe not need here, but best to add, then no need to consider that thing.

cosmosDB/src/test/java/org/talend/components/cosmosDB/CosmosDBOutputTestIT.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Test
+    public void outputTest() {
+        config.setAutoIDGeneration(true);
+
+        CosmosDBOutput cosmosDBOutput = new CosmosDBOutput(config, service, i18n);
+        cosmosDBOutput.init();
+        Record record = createData(1).get(0);
+        System.out.println(record);
```
wwang-talend: use assert?

pom.xml
```diff
@@ -53,6 +53,7 @@
         <module>extensions</module>
         <module>pubsub</module>
         <module>netsuite</module>
+		<module>cosmosDB</module>
```
wwang-talend: format?

pyzhou-talend:
@wwang-talend modified according to your comment, please help to review again.

PR #369, opened 2020-03-24
Wwang talend/tdi 43718

wwang-talend:


mongodb/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    <dependency>
+      <groupId>org.mongodb</groupId>
+      <artifactId>mongo-java-driver</artifactId>
+      <version>${mongodb.client.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.talend.components</groupId>
+      <artifactId>stream-json</artifactId>
+      <version>${stream-json.version}</version>
```
jzhao-talend, line 38: refer to other connector. better use ${project.version} ?

wwang-talend: thanks, fixed

wwang-talend:
the test failed as some other reason, not current connector

PR #372, opened 2020-03-25
fix(TDI-43896): Error while adding destination: r.map is not a function

wwang-talend:


wwang-talend:
https://talend.github.io/component-runtime/main/1.0.5/tutorial-handle-talend-component-migration.html

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
```diff
@@ -156,4 +162,30 @@ public void setKeys(List<String> keys) {
...
     }
 
+    @Slf4j
+    public static class Migration implements MigrationHandler {
+
+        @Override
+        public Map<String, String> migrate(int incomingVersion, Map<String, String> incomingData) {
+            log.debug("Starting JDBC sink component migration");
+
+            if (incomingVersion == 1) {
```
wwang-talend: the number is right? will check it

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
```diff
@@ -38,6 +43,7 @@
 import static org.talend.sdk.component.api.configuration.condition.ActiveIfs.Operator.AND;
 import static org.talend.sdk.component.api.configuration.condition.ActiveIfs.Operator.OR;
 
+@Version(value = 1, migrationHandler = OutputConfig.Migration.class)
```
rdubois: The version 1 is the current one. With the breaking change, you should bump to 2.

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
rdubois: Yes that version is correct.

When a component has the `@Version` tag, it defaults to `1`. When the `@Version` is missing, on TPD side, we also send the version `1` as a fallback.

jdbc/src/main/java/org/talend/components/jdbc/configuration/OutputConfig.java
```diff
@@ -156,4 +162,30 @@ public void setKeys(List<String> keys) {
...
+            log.debug("Starting JDBC sink component migration");
+
+            if (incomingVersion == 1) {
+                String old_property_path = "configuration.outputConfig.keys";
+                String new_property_path = "configuration.outputConfig.keys.keys";
+                String value = incomingData.get(old_property_path);
+
+                // as the new ui have been released, so need to do the special process for not breaking the right data
+                if (needMigration(value)) {
+                    incomingData.put(new_property_path, value);
```
rdubois: What about removing the original path from the map ? Without doing this, the component properties will be made of both:

```
configuration.outputConfig.keys -> [a, b, c]
configuration.outputConfig.keys.keys -> [a, b, c]
```
I would suggest to clean by removing the first one from the Map.

wwang-talend: i didn't remove it as i don't make sure if "configuration.outputConfig.keys" have meaning or value, for example :
`"configuration": {
      "useInvalidCharReplacement": false,
      "vacuum": true,
      "$maxBatchSize": 1000,
      "outputConfig": {
        "ignoreUpdate": [],
        "distributionKeys": [],
        "$configuration.outputConfig.actionOnData_name": "INSERT",
        "createTableIfNotExists": true,
        "rewriteBatchedStatements": true,
        "keys": ["email"],
        "keys": {"keys" : ["email"]},
        "sortStrategy": "COMPOUND",
        "varcharLength": -1,
        "actionOnData": "UPSERT",
        "distributionStrategy": "AUTO",
        "sortKeys": [],`

wwang-talend:
TODO : add junit

wwang-talend:
https://github.com/Talend/cloud-components/pull/320 jdbc-ee one

PR #398, opened 2020-04-16
feat(TDI-44040) : HTML type added

clesaec:
HTML type added in read file.


common-stream-io/stream-excel/src/main/java/org/talend/components/common/stream/input/excel/ExcelReader.java
```diff
@@ -0,0 +1,77 @@
...
+
+            if (headers >= 1) {
+                this.parseHeaderRow(rowIterator, headers);
+            }
+
+            return IteratorComposer.of(rowIterator).skipFooter(configuration.calcFooter())
+                    .map((Row row) -> this.toRecord.toRecord(row)).closeable(currentWorkBook).build();
+        } catch (IOException exIO) {
+            log.error("Error while reading excel input", exIO);
+            throw new UncheckedIOException("Error while reading excel input", exIO);
```
undx, line 57: i18n as was discussed yesterday ;-)

common-stream-io/stream-excel/src/main/java/org/talend/components/common/stream/input/excel/ExcelRecordReader.java
```diff
@@ -40,65 +31,25 @@
...
-                rows.next();
-            }
-            final Row headerRow = rows.next();
-            return this.toRecord.inferSchema(headerRow, true);
-        }
-        return null;
-    }
-
     @Override
     public void close() {
```
undx, line 53: remove empty method?

common/src/main/java/org/talend/components/common/text/SchemaUtils.java
```diff
@@ -0,0 +1,64 @@
...
+        return getUniqueName(result, previousNames);
+    }
+
+    private static String getUniqueName(String name, Set<String> previousNames) {
+        int index = 0;
+        String currentName = name;
+        while (previousNames.contains(currentName)) {
+            currentName = name + (++index);
+        }
+        return currentName;
```
undx, line 62: already seen this code in the connectors-se code base. Maybe, later factorize to an uniq place...

common-stream-io/stream-excel/src/main/java/org/talend/components/common/stream/input/excel/ExcelRecordReader.java
clesaec: No, view with Yves, we keep abstract method in interface (no default) to oblige to think about a potentially close method.

common/src/main/java/org/talend/components/common/text/SchemaUtils.java
clesaec: Yes !! Here, it's move in "common" place :)

PR #402, opened 2020-04-22
fix(TDI-44090): Update FileIO to improve columns name

wwang-talend:


pubsub/src/test/java/org/talend/components/pubsub/output/message/TextMessageGeneratorTest.java
```diff
@@ -40,7 +40,7 @@ public void init() {
         beanUnderTest = new TextMessageGenerator();
         beanUnderTest.setI18nMessage(new InternationalizationServiceFactory(() -> Locale.US).create(I18nMessage.class,
                 Thread.currentThread().getContextClassLoader()));
-        beanUnderTest.setRecordService(new RecordServiceImpl(null, null));
```
wwang-talend, line 43: when upgrade to 1.1.20-SNAPSHOT, meet a compiler issue which is different with 1.1.15, fix it like this

pom.xml
```diff
@@ -88,7 +88,8 @@
 
         <locales.version>[1.8,)</locales.version>
 
-        <component-runtime.version>1.1.15</component-runtime.version>
```
wwang-talend, line 91: don't merge as this pr is for test firstly, later will adjust 1.1.20-SNAPSHOT to 1.1.20. And adjust connector-se too which align with cloud-components

wwang-talend:
sure test in jenkins will failed as depend on Talend/component-runtime#428

pubsub/src/test/java/org/talend/components/pubsub/output/message/TextMessageGeneratorTest.java
wwang-talend: that is not because. :  https://github.com/Talend/component-runtime/pull/428/files , some other commit adjust it

wwang-talend:
close as not necessary now

PR #404, opened 2020-04-27
TDI-43732 : flatten CSV

rlecomte-talend:
During generation of a CSV message, all arrays and inner records are now omitted.

### DIFF
--- a/pubsub/src/main/java/org/talend/components/pubsub/output/message/CSVMessageGenerator.java
+++ b/pubsub/src/main/java/org/talend/components/pubsub/output/message/CSVMessageGenerator.java
@@ -108,79 +108,56 @@ public class CSVMessageGenerator extends MessageGenerator {
 
         @Override
         public void onBooleanArray(Schema.Entry entry, Optional<Collection<Boolean>> array) {
-            if (array.isPresent()) {
-                onArray(array.get());
-            } else {
-                csv.append(fieldDelimiter);
-            }
+
         }
 
         @Override
         public void onBytesArray(Schema.Entry entry, Optional<Collection<byte[]>> array) {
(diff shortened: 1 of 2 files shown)


pubsub/src/main/java/org/talend/components/pubsub/output/message/CSVMessageGenerator.java
```diff
@@ -108,79 +108,56 @@ public void onDatetime(Schema.Entry entry, Optional<ZonedDateTime> dateTime) {
 
         @Override
         public void onBooleanArray(Schema.Entry entry, Optional<Collection<Boolean>> array) {
-            if (array.isPresent()) {
-                onArray(array.get());
-            } else {
-                csv.append(fieldDelimiter);
-            }
+
```
clesaec, line 111: As RecordVisitor interface already defined a default for this method that do no op, you don't need to declare it at all

pubsub/src/main/java/org/talend/components/pubsub/output/message/CSVMessageGenerator.java
```diff
@@ -108,79 +108,56 @@ public void onDatetime(Schema.Entry entry, Optional<ZonedDateTime> dateTime) {
...
+        }
+
+        @Override
+        public RecordVisitor<String> onRecordArray(Schema.Entry entry, Optional<Collection<Record>> array) {
+            return this;
+        }
+
+        @Override
+        public RecordVisitor<String> onRecord(Schema.Entry entry, Optional<Record> record) {
+            return this;
```
clesaec, line 160: this tow methods (onRecordArray & onRecord) are defined the same way as default on interface. So, you don't need to declare it.

PR #405, opened 2020-04-28
fix(TDI-44110):Correct dependence and errror message.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-44110


wwang-talend:
netsuite build failed? not cause by current fix? but before failed reason is a more common random fail, but this time, why netsuite failed?

PR #406, opened 2020-04-28
Clesaec/tdi 44088 jdbc tests

clesaec:
JUnit test modify to upgrade perf : from 12 minute to 2 min 15 sec


jdbc/src/main/java/org/talend/components/jdbc/output/platforms/OraclePlatform.java
```diff
@@ -90,6 +92,16 @@ private String createColumn(final Column column) {
         ;
     }
 
+    @Override
+    protected String pkConstraintName(String table, List<Column> primaryKeys) {
+        // Avoid 'ORA-00972: identifier is too long' that limit oracle id to 30 chars max
+        final String name = super.pkConstraintName(table, primaryKeys);
+        if (name == null || name.length() <= 30) {
+            return name;
```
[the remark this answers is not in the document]
clesaec: Very general case, so it can be null, yes.

jdbc/src/test/java/org/talend/components/jdbc/suite/AbstractBaseJDBC.java
```diff
@@ -10,107 +10,86 @@
...
-    public String getTestTableName(final TestInfo info) {
-        return info.getTestClass().map(Class::getSimpleName).map(name -> name.substring(0, Math.min(5, name.length())))
-                .orElse("TEST") + "_"
-                + info.getTestMethod().map(Method::getName).map(name -> name.substring(0, Math.min(10, name.length())))
-                        .orElse("TABLE");
+    @Injected
+    private BaseComponentsHandler componentsHandler;
+
+    public I18nMessage getI18nMessage() {
+        return i18nMessage;
```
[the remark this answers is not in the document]
clesaec: Fields are injected by introspection, so they can't be final but should not be modify; so they are private.
Field getters could be protected instead of public. You think it should be change ?

jdbc/src/main/java/org/talend/components/jdbc/output/platforms/OraclePlatform.java
```diff
@@ -90,6 +92,16 @@ private String createColumn(final Column column) {
         ;
     }
 
+    @Override
+    protected String pkConstraintName(String table, List<Column> primaryKeys) {
```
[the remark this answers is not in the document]
clesaec: That's because in [test config](https://github.com/Talend/connectors-se/blob/clesaec/TDI-44088_JDBC_Tests/jdbc/src/test/resources/testcontainers.properties#L2), we use [oracleinanutshell/oracle-xe-11g](https://hub.docker.com/r/oracleinanutshell/oracle-xe-11g/tags), for 12c, the only docker hub image i found is [tqldtqmapirails/oracle-xe-12c](https://hub.docker.com/r/tqldtqmapirails/oracle-xe-12c/tags) but its size is **3.5GB** (compare to 0.6GB for 11g); so unit test run with 11g (too long with 12c).

ypiel-talend:
The number of jdbc UT in master and this branch are not the same :
- on master : https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/master/267/testReport/
- on this branch : https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/clesaec%2FTDI-44088_JDBC_Tests/13/testReport/

I just want to be sure that the remove of jdbc/src/test/java/org/talend/components/jdbc/WithDatabasesEnvironments.java and some test suite class like jdbc/src/test/java/org/talend/components/jdbc/testsuite/UIActionServiceTest.java
keep the same scope.

clesaec:
On test result of master, you can see that [oracle test](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/master/267/testReport/org.talend.components.jdbc.testsuite/UIActionServiceTest/getTableColumnFromDatabase_TestInfo__JdbcTestContainer__5_/) are skip, pb i solved, and [snowflake](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/master/267/testReport/org.talend.components.jdbc.testsuite/UIActionServiceTest/getTableColumnFromDatabase_TestInfo__JdbcTestContainer__7_/) also because there are voluntary [deactivate](https://github.com/Talend/connectors-se/blob/master/jdbc/src/test/java/org/talend/components/jdbc/testsuite/UIActionServiceTest.java#L52).
So, for examples on [UIActionServiceTest](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/master/267/testReport/org.talend.components.jdbc.testsuite/UIActionServiceTest/) master show 7 database tests for each method but execute really 5 (so you see 63 tests run, in fact it's (63/7)*5 = 45 tests really run compare to [test branch](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/clesaec%2FTDI-44088_JDBC_Tests/13/testReport/org.talend.components.jdbc.suite/JDBCBaseContainerTest$UIActionServiceTest/) there's 54 tests really passed (9 function for 6 database).
And it's the same for all suite test (divide by 7 to count test methods and compare with branch and divide by 6 give the same result, but on master, it's only 5/7 tests really passed). Later, we should see if there's a solution for snowflake as there's no testcontainers for this db.

jdbc/src/main/java/org/talend/components/jdbc/output/platforms/Platform.java
```diff
@@ -84,16 +86,21 @@ public String identifier(final String name) {
...
+                : ", CONSTRAINT " + pkConstraintName(metaData, table, primaryKeys) + " PRIMARY KEY "
                         + primaryKeys.stream().map(Column::getName).map(this::identifier).collect(joining(",", "(", ")"));
     }
 
-    private String pkConstraintName(String table, List<Column> primaryKeys) {
+    protected String pkConstraintName(DatabaseMetaData metaData, String table, List<Column> primaryKeys) throws SQLException {
         final String uuid = UUID.randomUUID().toString();
-        return "pk_" + table + "_" + primaryKeys.stream().map(Column::getName).collect(joining("_")) + "_"
+        final int nameLength = metaData.getMaxColumnNameLength();
+        String constraint = "pk_" + table + "_" + primaryKeys.stream().map(Column::getName).collect(joining("_")) + "_"
```
ypiel-talend, line 98: What do you think to use a digest algorithm on 'constraint' variable ? 
'uuid.substring(0, Math.min(4, uuid.length())' So no need  to do substring on uuid

clesaec: I'm not a fan of this pk name construction to. The idea is to generate a unique name, but it's not guaranteed. That's why i let it as it was (just add a substring to conform to database constraint).

The best way to respect uniqueness should just use uuid, but it would generate a non explicit name. It would also be the issue with a digest algorithm.
So, the pb is to keep explicit name and reduce pb of collision :
We may replace uuid + substring(0,4) to a random string of 4 char and start with name "XXXX_pk_<table>_<key cols> : but it still probability pb (out of scope of this PR ?)

ypiel-talend: I agree that it is out of scope of this ticket.
My idea was to define an explicit name as it is, but digest what is after pk_  when it is too long. So you loose explicit name only when it is too long.
But yes please create a work item ticket to change it after.

ypiel-talend: Thanks : https://jira.talendforge.org/browse/TDI-44208

PR #408, opened 2020-04-30
fix(TDI-44156) authorize spaces in report name

clesaec:
https://jira.talendforge.org/browse/TDI-44156

Before :
  - Didn't encode url 
Now :
  - Encode URL
 - Take into account that workday transform space in underscore for report name.


workday/src/main/java/org/talend/components/workday/service/AccessTokenConfigurer.java
```diff
@@ -24,6 +25,6 @@
     public void configure(Connection connection, ConfigurerConfiguration configuration) {
         log.debug("[configure] [{}] {}", connection.getMethod(), connection.getUrl());
         connection //
-                .withHeader("Content-Type", CONTENT_TYPE_APPLICATION_X_WWW_FORM_URLENCODED).withoutFollowRedirects();
+                .withHeader("Content-Type", CONTENT_TYPE_APPLICATION_X_WWW_FORM_URLENCODED); // .withoutFollowRedirects();
```
rlecomte-talend, line 28: If the code becomes useless, remove it instead of commenting it.

workday/src/main/java/org/talend/components/workday/dataset/QueryHelper.java
```diff
@@ -23,4 +28,12 @@
 
     Map<String, String> extractQueryParam();
 
+    default String encodeString(final String raw) {
+        try {
+            return URLEncoder.encode(raw, StandardCharsets.UTF_8.toString());
+        } catch (UnsupportedEncodingException ex) {
+            throw new WorkdayException("Encoding error for '" + raw + "'", ex);
```
ypiel-talend, line 35: Precise that it concern url encoding.

workday/src/main/java/org/talend/components/workday/service/AccessTokenConfigurer.java
ypiel-talend, line 28: Why remove withoutFollowRedirects() ? In which way the issue impact the redirections ?

workday/src/main/java/org/talend/components/workday/service/AccessTokenService.java
```diff
@@ -35,7 +35,7 @@ public Token findToken(WorkdayDataStore datastore) {
     }
 
     private boolean isTokenTooOld() {
-        return token.getExpireDate().isAfter(Instant.now().minus(30, ChronoUnit.SECONDS));
+        return token.getExpireDate().isBefore(Instant.now().plus(30, ChronoUnit.SECONDS));
```
ypiel-talend, line 38: Is it part of the initial issue about url encoding ?

clesaec: It's an old 'not blocking' bug; so, this correction is the opportunity to correct it.

workday/src/main/java/org/talend/components/workday/dataset/QueryHelper.java
clesaec: done

workday/src/main/java/org/talend/components/workday/service/AccessTokenConfigurer.java
clesaec: Was because the pipeline designer instance i downloaded still didn't support this method (tck version); i put it again

clesaec: code un-commented.

PR #421, opened 2020-05-20
Rlecomte/tdi 44222 ftp connection pool

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-44222
Added a connection pool in FTPService to reduce the number of created connections.
Some Unit test are disabled because of an issue with permissions checking. This will be resolved in another PR.


ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
```diff
@@ -99,7 +107,22 @@ public String generateFilePath(String dirPath, String filename) {
         return dirPath + (dirPath.endsWith(PATH_SEPARATOR) ? "" : PATH_SEPARATOR) + filename;
     }
 
-    public GenericFTPClient getClient(FTPDataStore dataStore) {
+    public GenericFTPClient getClient(FTPConnectorConfiguration configuration) {
+
+        if (clientMap == null) {
+            clientMap = Collections.synchronizedMap(new HashMap<>());
```
clesaec, line 113: Use class java.util.concurrent.ConcurrentHashMap
=> better optimized for conccurence.

ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
```diff
@@ -99,7 +107,22 @@ public String generateFilePath(String dirPath, String filename) {
...
+        }
+
+        GenericFTPClient client = clientMap.get(configuration.getConfigKey());
+        if (client == null || !client.isConnected()) {
+            client = getClient(configuration.getDataSet().getDatastore());
+            clientMap.put(configuration.getConfigKey(), client);
+        }
```
clesaec, line 120: Use Map.compute(K key,
            BiFunction<? super K, ? super V, ? extends V> remappingFunction) here.
remapping function could simply be a private function :
```
private GenericFTPClient computeClient(String key, GenericFTPClient current) {
       if (current == null || !current.isConnected()) {
            return  getClient(configuration.getDataSet().getDatastore());
       }
       return current
}
```
and call `GenericFTPClient client = clientMap.compute(configuration.getConfigKey(), this::computeClient);`
So, it respect multi-threading contract

ftp/src/test/java/org/talend/components/ftp/output/FTPOutputTest.java
```diff
@@ -93,6 +96,39 @@ void buildConfig() {
         configuration.setDebug(true);
     }
 
+    // @EnvironmentalTest
```
clesaec, line 99: Can't be active ?

ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
rlecomte-talend: I did that on the first occurence, but I had to add the isConnected test.

clesaec: My computeClient example function also has "isConnected" test

rlecomte-talend: OK, done

PR #423, opened 2020-05-22
fix(TDI-44233): add dependence of httpclient

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-44233

### DIFF
--- a/cosmosDB/pom.xml
+++ b/cosmosDB/pom.xml
@@ -34,18 +34,22 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
+      <scope>compile</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


cosmosDB/pom.xml
```diff
@@ -69,30 +62,6 @@
 
   <build>
     <plugins>
-      <plugin>
```
wwang-talend, line 72: why remove this, no meaning?

pom.xml
```diff
@@ -108,6 +108,8 @@
 
         <commons-csv.version>1.4</commons-csv.version>
         <guava.version>20.0</guava.version>
+		<httpclient.version>4.5.10</httpclient.version>
```
wwang-talend, line 111: best to correct format in future

cosmosDB/pom.xml
pyzhou-talend: Yes. Seems added by mistake before.

PR #427, opened 2020-05-29
fix(TDI-44208) : change pk generation

clesaec:
Change pk name generation in case of overflow.
JIRA : https://jira.talendforge.org/browse/TDI-44208


clesaec:
This method is already test by suite with different container.
But right, this class may not respect [single responsibility principle](http://blogs.developpeur.org/tja/archive/2010/01/08/architecture-partie-1-srp-single-responsability-principle.aspx), so it can't be test directly. May sql request construction and sql execution should be separated, but it's over of this PR goal.

PR #429, opened 2020-06-03
fix(TDI-44314) : workday with tck records

clesaec:
Correction for 
- https://jira.talendforge.org/browse/TDI-44314
- https://jira.talendforge.org/browse/TDI-44316
Sampling does not support schema variation.

### DIFF
--- a/common/src/test/java/org/talend/components/common/schema/RecordGuessWorkTest.java
+++ b/common/src/test/java/org/talend/components/common/schema/RecordGuessWorkTest.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright (C) 2006-2020 Talend Inc. - www.talend.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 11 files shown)


common/src/test/java/org/talend/components/common/schema/RecordGuessWorkTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+        JsonToRecord tr = new JsonToRecord(this.factory);
+        final Record r1 = tr.toRecord(jsonObject1, schema);
+        final Record r2 = tr.toRecord(jsonObject2, schema);
+        Assertions.assertNotNull(r1);
+        Assertions.assertNotNull(r2);
+    }
+
+    @Test
+    void loadComplexFromFilesRecordsA(){
```
ypiel-talend, line 223: This unit test fails

clesaec: Yes, correction done

PR #436, opened 2020-06-18
Rlecomte/tdi 4422 ftp connection pool

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-44222

Issue found during final check


ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
```diff
@@ -151,9 +151,11 @@ public void releaseClient(FTPConnectorConfiguration configuration) {
...
         List<GenericFTPFile> files = ftpClient.listFiles(path);
         String[] pathElements = path.split(PATH_SEPARATOR);
-        String pathLastElement = pathElements[pathElements.length - 1];
-        if (files.size() == 1) {
-            return !files.get(0).isDirectory() && (files.get(0).getName().equals(pathLastElement));
+        if (pathElements.length >= 1) {
+            String pathLastElement = pathElements[pathElements.length - 1];
+            if (files.size() == 1) {
+                return !files.get(0).isDirectory() && (files.get(0).getName().equals(pathLastElement));
+            }
```
clesaec, line 158: inverse tests :
```
if (files.size() == 1) {
    String[] pathElements = path.split(PATH_SEPARATOR);        
    if (pathElements.length >= 1) {
           String pathLastElement = pathElements[pathElements.length - 1];
           return !files.get(0).isDirectory() && (files.get(0).getName().equals(pathLastElement));
```
So, you won't need always call costly 'split' operator.
or even better (without split) :
```
String pathLastElement = path.substring(path.lastIndexOf("/") + 1);
```

rlecomte-talend: Done

ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
```diff
@@ -150,9 +150,8 @@ public void releaseClient(FTPConnectorConfiguration configuration) {
      */
     public boolean pathIsFile(GenericFTPClient ftpClient, String path) {
         List<GenericFTPFile> files = ftpClient.listFiles(path);
-        String[] pathElements = path.split(PATH_SEPARATOR);
-        String pathLastElement = pathElements[pathElements.length - 1];
         if (files.size() == 1) {
+            String pathLastElement = path.substring(path.lastIndexOf("/") + 1);
             return !files.get(0).isDirectory() && (files.get(0).getName().equals(pathLastElement));
```
ypiel-talend, line 155: "!files.get(0).isDirectory()" is not enough to know if it is a file or not ?

rlecomte-talend: Nope. Is the target is a directory, list will return its content, so files.get(0) will then be the first item in the directory and not the directory itself. If the target is a file, list will return the file itself.

ftp/src/main/java/org/talend/components/ftp/service/FTPService.java
```diff
@@ -80,11 +80,15 @@ public SuggestionValues suggestPath(@Option final FTPDataStore datastore, @Optio
...
+        if (PATH_SEPARATOR.equals(path)) {
+            return true;
+        }
+
+        if (path.endsWith(PATH_SEPARATOR) && path.length() > PATH_SEPARATOR.length()) {
             path = path.substring(0, path.length() - PATH_SEPARATOR.length());
         }
-        String[] pathElements = path.split(PATH_SEPARATOR);
-        String pathLastElement = pathElements[pathElements.length - 1];
+        String pathLastElement = path.substring(path.lastIndexOf(PATH_SEPARATOR));
```
clesaec, line 91: `final String pathLastElement = path.substring(path.lastIndexOf(PATH_SEPARATOR)  + PATH_SEPARATOR.length() );`

should better work (try step by step)

rlecomte-talend: done

ftp/src/main/java/org/talend/components/ftp/service/ftpclient/ApacheFTPClient.java
```diff
@@ -153,18 +160,44 @@ public void enableDebug(Logger alog) {
...
+        }
+
+        try {
+            ftpClient.setListHiddenFiles(true);
+            Arrays.stream(ftpClient.listFiles(path)).forEach(System.out::println);
+            FTPFile[] candidates = ftpClient.listFiles(path, f -> f.getName().equals("."));
+            if (candidates.length == 0) {
+                // No way to list directory itself, need to go to parent
+                String dirName = path.substring(path.lastIndexOf('/') + 1);
+                Arrays.stream(ftpClient.listFiles(path + "/..")).forEach(System.out::println);
```
clesaec, line 176: Is System.out.println is volontary or for debug

ftp/src/main/java/org/talend/components/ftp/service/ftpclient/ApacheFTPClient.java
```diff
@@ -153,18 +160,44 @@ public void enableDebug(Logger alog) {
...
+                candidates = ftpClient.listFiles(path + "/..", f -> f.getName().equals(dirName) || f.getName().equals(path));
+            }
+            ftpClient.setListHiddenFiles(false);
+            return toGenericFTPFile(candidates[0]).isWritable();
```
clesaec, line 180: candidates length can still be zero ... in case, return false ?

ftp/src/main/java/org/talend/components/ftp/service/ftpclient/ApacheFTPClient.java
rlecomte-talend: removed

ftp/src/main/java/org/talend/components/ftp/service/ftpclient/ApacheFTPClient.java
rlecomte-talend: Test added, but in this case, it returns true. If there's no way know if folder is writable, we can't block the pipeline.

rlecomte-talend:
FTP connector is no more in connectors-se

PR #438, opened 2020-06-29
fix(TDI-44403): Fix component names

dmytro-sylaiev:


mongodb/src/main/resources/org/talend/components/mongodb/sink/Messages.properties
```diff
@@ -1,4 +1,4 @@
-MongoDB.Sink._displayName=MongoDB sink
+MongoDB.Sink._displayName=MongoDBOutput
```
ypiel-talend, line 1: Should be : "MongoDB Output" with a space

mongodb/src/main/resources/org/talend/components/mongodb/source/Messages.properties
```diff
@@ -1,5 +1,5 @@
-MongoDB.CollectionQuerySource._displayName=MongoDB collection query source
-MongoDB.CollectionSource._displayName=MongoDB collection source
+MongoDB.CollectionQuerySource._displayName=MongoDBInputQuery
+MongoDB.CollectionSource._displayName=MongoDBInput
```
ypiel-talend, line 2: Should be "MongoDB Input" with a space.

ypiel-talend: Shoud be "MongoDB Input Query" with spaces.

PR #447, opened 2020-07-21
Pyzhou/tdi 44170 fix cosmos junit

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-44170


cosmosDB/src/main/java/org/talend/components/cosmosDB/output/CosmosDBOutput.java
```diff
@@ -109,7 +109,6 @@ private void createDocumentCollectionIfNotExists() {
                 // flexibility including string range queries.
                 RangeIndex index = new RangeIndex(DataType.String);
                 index.setPrecision(-1);
-                IndexingPolicy indexingPolicy = new IndexingPolicy();
```
wwang-talend, line 112: where to use it?

pyzhou-talend: Nowhere, So I dorp it.

PR #453, opened 2020-08-06
fix(TDI-44667) : fix npe with null value when generate csv

ypiel-talend:
https://jira.talendforge.org/browse/TDI-44667


Voievidko (approved):
LGTM

PR #456, opened 2020-08-19
Ypiel/tdi 44479 rest check connection

ypiel-talend:
https://jira.talendforge.org/browse/TDI-44479

Check connection was done on url base, like : https://mydomain.com/api/v1 that fails.
Now, we only check connection on domain https://mydomain.com.


Voievidko (approved):
LGTM

Voievidko (dismissed):
LGTM

Voievidko (approved):

PR #457, opened 2020-08-24
Rlecomte/tdi 44729 move ftp to connectors ee

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-44729

FTP connector removed from connectors-se


ypiel-talend (approved):
LGTM

PR #458, opened 2020-08-26
fix(TDI-44750): add null fields in record's schema with json

undx:
https://jira.talendforge.org/browse/TDI-44750

When returning a Record from a Json file with null values, the null values are skipped from schema. 
So, the record's schema is not fully feeded.

### DIFF
--- a/azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/json/JsonConverter.java
+++ b/azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/json/JsonConverter.java
@@ -12,6 +12,8 @@
  */
 package org.talend.components.adlsgen2.common.format.json;
 
+import static java.util.stream.Collectors.toList;
+
 import java.io.Serializable;
 import java.time.ZonedDateTime;
 import java.util.Optional;
@@ -34,8 +36,6 @@ import org.talend.sdk.component.api.service.record.RecordBuilderFactory;
 
 import lombok.extern.slf4j.Slf4j;
 
-import static java.util.stream.Collectors.toList;
-
 @Slf4j
 public class JsonConverter implements RecordConverter<JsonObject>, Serializable {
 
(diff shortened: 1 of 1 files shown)


rlecomte-talend (approved):
LGTM

azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/json/JsonConverter.java
```diff
@@ -68,10 +66,7 @@ public Schema inferSchema(final JsonObject record) {
 
     @Override
     public Record toRecord(final JsonObject record) {
-        if (schema == null || record.keySet().size() > schema.getEntries().size()) {
-            schema = inferSchema(record);
-        }
-        return convertJsonObjectToRecord(schema, record);
+        return convertJsonObjectToRecord(inferSchema(record), record);
```
clesaec, line 69: rename convertJsonObjectToRecord in toRecord and make it public, it should be used.

azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/json/JsonConverter.java
```diff
@@ -34,8 +36,6 @@
 
 import lombok.extern.slf4j.Slf4j;
 
-import static java.util.stream.Collectors.toList;
-
 @Slf4j
 public class JsonConverter implements RecordConverter<JsonObject>, Serializable {
```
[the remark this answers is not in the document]
undx: This is a fix issue, not a refactoring one... A work item should be created for that.

azure-dls-gen2/src/main/java/org/talend/components/adlsgen2/common/format/json/JsonConverter.java
undx: don't get the point here.

clesaec (approved):
:+1:

undx:
obsolete

PR #461, opened 2020-09-02
feat(TDI-44719):Integrate Cosmos to TFD

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-44719


cosmosDB/src/main/java/org/talend/components/cosmosDB/input/CosmosDBInput.java
```diff
@@ -45,7 +48,7 @@
 
     private final RecordBuilderFactory builderFactory;
 
-    private transient CosmosDBService service;
```
wwang-talend, line 48: why remove "transient"? only a question, seems the service is only created at runtime, no need to ser/deser through network

wwang-talend (approved):
LGTM

cosmosDB/src/main/java/org/talend/components/cosmosDB/input/CosmosDBInput.java
pyzhou-talend: According to my test. It was created when the Input/Output instance was created and then distribute to the work note. There will be an NPE with the "transient" exist.

PR #467, opened 2020-09-10
prepare(release/1.13.0): set locales to 1.13.1 for deterministic release

undx:
- also correct bad formatting in jdbc


rlecomte-talend (approved):
LGTM

PR #472, opened 2020-10-08
feat(TDI-44986): Add aws-common module

chmyga:
Add common aws module with authentication services. https://jira.talendforge.org/browse/TDI-44986


dmytro-sylaiev (approved):

chmyga:
It was decided to move the module to connectors-ee

PR #473, opened 2020-10-10
fix(TFD-11651): add the epoll native 4.1.48-Final to pubsub component

rdubois:
A wrong version of this epoll dependency was used in grpc-netty, and it led to class cast exception. See the JIRA comment for more details.

### DIFF
--- a/pubsub/pom.xml
+++ b/pubsub/pom.xml
@@ -42,6 +42,11 @@
             <version>2.0.26.Final</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


wwang-talend (approved):

wwang-talend:
the issue only appear in pubsub? look a generic issue, also need to align it for other connectors?

rdubois:
Good question @wwang-talend, I didn't really check other components. This investigation has been done because of the TPD QA found out an issue with this connnector only.

PR #477, opened 2020-10-22
feat(TDI44972) : add fixed parameters

ypiel-talend:
https://jira.talendforge.org/browse/TDI-44972

**New behavior:**
Add new available property for jdbc handler : fixed_parameters
Which is a list of key/value that will be added as jdbc url parameter automatically.


OleksiiNimych (approved):
LGTM

PR #478, opened 2020-10-23
feat(TDI-44723): BigQuery TCK append and override

OleksiiNimych:
**What is the current behavior?** (You can also link to an open issue here)
https://jira.talendforge.org/browse/TDI-44723

**What is the new behavior?**
Added possibility to truncate BigQuery table on pipeline designer. Append option was available before.

### DIFF
--- a/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
+++ b/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
@@ -64,7 +64,6 @@ public class BigQueryOutput implements Serializable {
 
     private BigQueryService service;
 
-    private JobId jobId;
 
     public BigQueryOutput(@Option("configuration") final BigQueryOutputConfig configuration, BigQueryService bigQueryService,
             I18nMessage i18n) {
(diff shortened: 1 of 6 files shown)


bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -73,6 +74,51 @@ public BigQueryOutput(@Option("configuration") final BigQueryOutputConfig config
...
+                            + configuration.getDataSet().getBqDataset() + "." + configuration.getDataSet().getTableName() + "`")
+                    .setUseLegacySql(false).build();
+            Job job = bigQuery.create(JobInfo.newBuilder(queryConfig).setJobId(getNewUniqueJobId(bigQuery)).build());
+            try {
+                job = job.waitFor();
+            } catch (InterruptedException e) {
+                log.warn(i18n.errorQueryExecution(), e);
+            }
+            if (job.isDone()) {
+                log.info("Truncate query successfully completed");
```
ypiel-talend, line 101: Should be i18n

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -64,6 +63,8 @@
 
     private transient TableId tableId;
 
+    private transient TableId tempTableId;
```
ypiel-talend, line 66: Those 'id' should be owned by a service. Indeed, connector can be serialized/deserialized between several node and or  garbaged and a new one created etc... services are stable on node, but it generates some other issue... Need to imply @undx

ypiel-talend (changes requested):
Need to imply @undx to discuss about solution.

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
ypiel-talend: A/ truncate is executed twice :truncate is done in connector's constructor : It will be done twice. Indeed, TCK will instantiate it first to transform to beam. Then after the 'orchestrator' will create an instance. This one will be serialized/de-serialized in all worker (in REG2 threads, but could be nodes). Not a big deal to have truncate done twice I think. But it is not very good practice to execute some processin constructor, but currently no other way (we think to a better approach but need some development in TFD / TCK).
B/ I think tempTableId UUID should not be transient and may be instantiated in @PostConstruct instead of @ElementListener. Indeed, post construct is called when the connector is deserialized in a worker (thread or node). If 'preDestroy' is not called, and we should not rely on it, data in tmp table will be lost. I think the INSERT INTO finalTable SELECT FROM tempTable should be done in afterGroup
C/ Why we don't directly insert in final table ? What is the need for this temp table mechanism ? (I didn't follow all the story)

PR #484, opened 2020-11-03
TDI-45000 : default name for excel sheet

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-45000

Added a default value for excel sheet.


clesaec (approved):
:+1:

dmytro-sylaiev (approved):

PR #488, opened 2020-11-12
feat(TDI-44873) : bump talend-mscrm to bump oliingo/odata version

ypiel-talend:
https://jira.talendforge.org/browse/TDI-44873


OleksiiNimych (approved):
LGTM

PR #492, opened 2020-11-20
TDI-45123 : migrate pubsub to connectors ee

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-45123

Removed pubsub from se


dmytro-sylaiev (approved):
Let the connector-se go :)

PR #493, opened 2020-11-20
feat(TDI-45226) : Clean REST check connection method

ypiel-talend:
https://jira.talendforge.org/browse/TDI-45226


OleksiiNimych (approved):
LGTM

PR #495, opened 2020-11-24
fix(TDI-45164): [CVE] org.apache.logging.log4j log4j-api in connectors-ee and cloud-components and align for se

wwang-talend:


vyu-talend (approved):
LGTM

PR #497, opened 2020-11-25
feat(TDI-45261) : Prevent REST connector to access internal api of a cloud engine

ypiel-talend:
https://jira.talendforge.org/browse/TDI-45261


common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -0,0 +1,74 @@
...
+public class ValidateSites {
+
+    public final static boolean CAN_ACCESS_LOCAL = Boolean.valueOf(System.getProperty("connectors.enable_local_network_access",
+            System.getenv().getOrDefault("CONNECTORS_ENABLE_LOCAL_NETWORK_ACCESS", "true")));
+
+    public final static boolean ENABLE_MULTICAST_ACCESS = Boolean
+            .valueOf(System.getProperty("connectors.enable_multicast_network_access",
+                    System.getenv().getOrDefault("CONNECTORS_ENABLE_MULTICAST_NETWORK_ACCESS", "true")));
+
+    private final static List<String> ADDITIONAL_LOCAL_HOSTS = Arrays.asList(new String[] { "224.0.0" // local multicast : from
```
coheigea, line 32: We should change 224.0.0 -> 224.0.0. - as otherwise it could match xxx.224.0.0 ?

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -0,0 +1,74 @@
...
+        return isValidSite(base, CAN_ACCESS_LOCAL, ENABLE_MULTICAST_ACCESS);
+    }
+
+    /**
+     * This method returns if the given url is valid depending of paremeter.
+     * We can make local sites and multicast class of addresses invalid.
+     *
+     * @param surl
+     * @param can_access_local
+     * @param enable_multicat_access
```
coheigea, line 49: enable_multicat_access -> enable_multicast_access

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -0,0 +1,74 @@
...
+                return false;
+            }
+
+            if (can_access_local) {
+                // we can access local sites
+                return true;
+            }
+
+            return !inetAddress.isSiteLocalAddress()
+                    && !ADDITIONAL_LOCAL_HOSTS.stream().filter(h -> host.contains(h)).findFirst().isPresent();
```
coheigea, line 69: Should we use isLoopbackAddress() here instead?

coheigea (changes requested):

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
rlecomte-talend: Done

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
rlecomte-talend: Done

coheigea (approved):

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ValidateSites {
+
+    public final static boolean CAN_ACCESS_LOCAL = Boolean.valueOf(System.getProperty("connectors.enable_local_network_access",
+            System.getenv().getOrDefault("CONNECTORS_ENABLE_LOCAL_NETWORK_ACCESS", "true")));
```
dmytro-sylaiev, line 28: if not specified, access to local API is allowed. Is it intended? So where does it specified then?

dmytro-sylaiev (approved):
LGTM if access is ON by default

coheigea:
Looks like there is a conflict that needs to be updated.

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
rlecomte-talend: Yes, that's the default behavior.

coheigea (approved):

dmytro-sylaiev (approved):
LGTM

PR #498, opened 2020-11-26
fix(TDI-44955) : retrieve polling parameters in instanciator.

ypiel-talend:
There were a NPE when using REST in streaming mode.
https://jira.talendforge.org/browse/TDI-44955


OleksiiNimych (approved):
LGTM

PR #503, opened 2020-12-01
feat(TDI-44723): add truncate option

OleksiiNimych:
https://jira.talendforge.org/browse/TDI-44723
Add truncate option to the bigquery connector

### DIFF
--- a/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
+++ b/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
@@ -195,7 +195,16 @@ public class BigQueryOutput implements Serializable {
         }
 
         if (BigQueryOutputConfig.TableOperation.TRUNCATE == configuration.getTableOperation()) {
-            loadData();
+            try {
+                loadData();
+            } catch (BigQueryException e) {
+                if (e.getMessage().contains("Already Exists")) {
+                    loadData();
+                } else {
+                    log.warn(e.getMessage());
+                }
+            }
+            deleteTempBlob();
         } else {
             streamData();
         }
@@ -260,6 +269,20 @@ public class BigQueryOutput implements Serializable {
             log.error(exIO.getMessage());
         }
 
+        JobInfo jobInfo = buildJobInfo();
+        Job job = bigQuery.create(jobInfo);
+        try {
+            job = job.waitFor();
+        } catch (InterruptedException e) {
+            log.warn(e.getMessage());
+        }
+        if (job.getStatus().getError() != null) {
+            log.warn("BigQuery was unable to load into the table due to an error:" + job.getStatus().getError());
+        }
+
+    }
+
+    private JobInfo buildJobInfo() {
         String sourceUri = "gs://" + configuration.getDataSet().getGsBucket() + "/" + blobName;
         Table table = bigQuery.getTable(tableId);
         Schema schema = table.getDefinition().getSchema();
@@ -280,15 +303,10 @@ public class BigQueryOutput implements Serializable {
             }
             jobInfo = JobInfo.of(loadConfigurationBuilder.build());
         }
-        Job job = bigQuery.create(jobInfo);
-        try {
-            job = job.waitFor();
-        } catch (InterruptedException e) {
-            log.warn(e.getMessage());
-        }
-        if (job.getStatus().getError() != null) {
-            log.warn("BigQuery was unable to load into the table due to an error:" + job.getStatus().getError());
-        }
+        return jobInfo;
+    }
+
+    private void deleteTempBlob() {
         storage.delete(blobInfo.getBlobId());
     }
 
--- a/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
+++ b/bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
@@ -89,8 +89,6 @@ public class BigQueryOutput implements Serializable {
 
     private transient RecordWriter recordWriter;
 
-    private transient String blobName;
-
     private transient BlobInfo blobInfo;
 
     private transient WriteChannel writer;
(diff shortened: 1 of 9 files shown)


rlecomte-talend (dismissed):
LGTM

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
...
+            jobInfo = JobInfo.of(loadConfigurationBuilder.build());
+        }
+        Job job = bigQuery.create(jobInfo);
+        try {
+            job = job.waitFor();
+        } catch (InterruptedException e) {
+            log.warn(e.getMessage());
+        }
+        if (job.getStatus().getError() != null) {
+            log.warn("BigQuery was unable to load into the table due to an error:" + job.getStatus().getError());
```
ypiel-talend: Work item to define a clear rule about what should append on such error ?
- stop pipeline ?
- Define a reject flow ?
- just log as here ?

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
...
+        }
+        storage.delete(blobInfo.getBlobId());
+    }
+
     private TableReference createTableReference() {
```
ypiel-talend, line 313: Remove if not used.

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutputConfig.java
```diff
@@ -58,6 +58,14 @@
...
          */
         CREATE_IF_NOT_EXISTS,
+        /**
+         * Specifies that write should replace a table.
+         *
+         * <p>
+         * If the output table does not exist, the write fails.
+         *
+         */
+        TRUNCATE,
```
ypiel-talend, line 68: Should be possible to have CREATE_IF_NOT_EXIST + TRUNCATE at the same time.
For instance bigquery config in studio:
image

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
ypiel-talend: warn message should be i18n

ypiel-talend (changes requested):
Some remarks

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
         }
     }
 
+    private void loadData() {
+        try {
+            this.recordWriter.add(records);
+            this.recordWriter.end();
```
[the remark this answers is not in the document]
OleksiiNimych: .close() call in the .end()
https://github.com/Talend/connectors-se/blob/master/common-stream-io/stream-api/src/main/java/org/talend/components/common/stream/api/output/RecordWriter.java#L28-L31

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
...
+        } catch (IOException exIO) {
+            log.error(exIO.getMessage());
```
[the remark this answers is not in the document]
OleksiiNimych: done

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -66,18 +79,78 @@
...
+    @BeforeGroup
+    public void beforeGroup() {
+        records = new ArrayList<>();
+        if (BigQueryOutputConfig.TableOperation.TRUNCATE == configuration.getTableOperation()) {
+            Blob blob = getNewBlob();
+            WriteChannel writer = blob.writer();
+            try {
+                recordWriter = buildWriter(writer);
+            } catch (IOException e) {
+                log.warn(e.getMessage());
```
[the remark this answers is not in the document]
OleksiiNimych: done

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
...
+        try {
+            this.recordWriter.add(records);
+            this.recordWriter.end();
+        } catch (IOException exIO) {
+            log.error(exIO.getMessage());
+        }
+
+        String sourceUri = "gs://" + configuration.getDataSet().getGsBucket() + "/" + blobName;
+        Table table = bigQuery.getTable(tableId);
+        Schema schema = table.getDefinition().getSchema();
```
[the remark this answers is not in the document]
OleksiiNimych: Updated to use tableSchema

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -173,6 +252,46 @@ public void afterGroup() {
...
+            if (!firstJob.isDone()) {
+                try {
+                    firstJob.waitFor();
+                } catch (InterruptedException e) {
+                    log.warn(e.getMessage());
+                }
+            }
+            jobInfo = JobInfo.of(loadConfigurationBuilder.build());
+        }
+        Job job = bigQuery.create(jobInfo);
```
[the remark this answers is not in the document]
OleksiiNimych: Added code that recall the method to load the data

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
OleksiiNimych: Removed

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
OleksiiNimych: Added i18n

ypiel-talend (approved):
LGTM

bigquery/src/main/java/org/talend/components/bigquery/output/BigQueryOutput.java
```diff
@@ -66,18 +79,107 @@
...
+        JobId jobId;
+        do {
+            jobId = JobId.of(UUID.randomUUID().toString());
+        } while (bigQuery.getJob(jobId) != null);
+        return jobId;
+    }
+
     @PostConstruct
     public void init() {
+        bigQuery = service.createClient(connection);
```
ypiel-talend, line 119: What was identify in previous code review is :
"Everything that can have side-effect on your treatment (ie : is part of your treatment) should not be in setup (@PostConstruct) nor teardown (@PreDestroy)."
Here it manages the truncate of the table, it can have side effect on the treatment so should not be in @PostConstruct.

You can move it in @BeforeGroup.

But it will be called more often and so 'bigQuery.getJob(jobId)' that can take some times !
To improve performance, we can rely on a service method with a local cache.
So, could you try to change 'bigQuery.getJob(jobId)' by something like 'service.isJobDone(jobId)'.
Use a LocalCache Service (built-in service of TCK) to check if there is a value for jobId key, 
if not, call  'bigQuery.getJob(jobId)'
and if not null store a boolean in the cache, so next call you will see in the cache thatthe job was done, and you will not have to call bigquery.

ypiel-talend (changes requested):
Need a change

ypiel-talend (approved):
LGTM

PR #507, opened 2020-12-03
chore(release) : skipTest on release due to new dockerhub policy.

ypiel-talend:
release must skip tests due to dockerhub new policy.


undx (approved):
LGTM

PR #518, opened 2020-12-13
feat(TDI-45199) : jsonToRecord shoud generate nullable fields

ypiel-talend:
https://jira.talendforge.org/browse/TDI-45199

### DIFF
--- a/common-stream-io/stream-json/src/main/java/org/talend/components/common/stream/input/json/JsonToRecord.java
+++ b/common-stream-io/stream-json/src/main/java/org/talend/components/common/stream/input/json/JsonToRecord.java
@@ -64,25 +64,26 @@ public class JsonToRecord {
             case ARRAY: {
                 final List<Object> items = value.asJsonArray().stream().map(this::mapJson).collect(toList());
                 builder.withArray(factory.newEntryBuilder().withName(key).withType(Schema.Type.ARRAY)
-                        .withElementSchema(getArrayElementSchema(factory, items)).build(), items);
+                        .withElementSchema(getArrayElementSchema(factory, items)).withNullable(true).build(), items);
                 break;
             }
             case OBJECT: {
                 final Record record = toRecord(value.asJsonObject());
                 builder.withRecord(factory.newEntryBuilder().withName(key).withType(Schema.Type.RECORD)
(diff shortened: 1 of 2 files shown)


common-stream-io/stream-json/src/main/java/org/talend/components/common/stream/input/json/JsonToRecord.java
```diff
@@ -64,25 +64,27 @@ public Record toRecord(final JsonObject object) {
             case ARRAY: {
                 final List<Object> items = value.asJsonArray().stream().map(this::mapJson).collect(toList());
                 builder.withArray(factory.newEntryBuilder().withName(key).withType(Schema.Type.ARRAY)
-                        .withElementSchema(getArrayElementSchema(factory, items)).build(), items);
+                        .withElementSchema(getArrayElementSchema(factory, items)).withNullable(true).build(), items);
```
clesaec, line 67: Should be better to start on line 65 with
`final Entry.Builder entryBuilder = this.factory.newEntryBuilder().withName(key).withNullable(true);`
Then work with it on each case; re-adapt Number.setNumber method with 
`public void setNumber(Record.Builder builder, Entry.Builder entryBuilder, JsonNumber number)`

And make `case STRING:` not a particular case.
So it will make less copy/past code.
What do you think ?

ypiel-talend: Done

common-stream-io/stream-json/src/main/java/org/talend/components/common/stream/input/json/JsonToRecord.java
```diff
@@ -203,8 +205,9 @@ public Number getNumber(JsonNumber number) {
                 return number.doubleValue();
             }
 
-            public void setNumber(Record.Builder builder, String key, JsonNumber number) {
-                builder.withDouble(key, number.doubleValue());
+            public void setNumber(Record.Builder builder, Schema.Entry.Builder entryBuilder, String key, JsonNumber number) {
```
clesaec, line 208: you don't need key parameter now (it included in entryBuilder), it can be removed to simplify method

clesaec (approved):
:+1:

PR #519, opened 2020-12-14
chore(TDI-45326): update talend mscrm lib

chmyga:
depends on commit https://github.com/Talend/tdi-studio-se/pull/5629


clesaec (commented):
:+1:

clesaec (approved):

PR #520, opened 2020-12-14
fix(TDI-45324): Upgrade POI version for azureblob

dmytro-sylaiev:


dmytro-sylaiev:
Currently there's a conflict with pipeline designer about commons-compress version which we can't resolve on our side.
Do not merge till resolution.

More information: https://jira.talendforge.org/browse/TDI-45324

PR #523, opened 2020-12-17
fix(TDI-45386): reduce azure-dls-gen2 dependencies

undx:
https://jira.talendforge.org/browse/TDI-45386

- reduce dependencies on hadoop


clesaec (approved):
:+1: 
Other needs on this connector, 
- use [azure lib ](https://mvnrepository.com/artifact/com.azure/azure-storage-file-datalake/12.3.0)
- All tests inherits from AdlsGen2TestBase, which depends on credentials; but lot of don't need it and could be used as Unit.
- Use of common-stream
-

undx:
As discussed in daily, this is not the point on this issue. This is not a code change.
Related issues should be created as work items.

PR #491, opened 2020-11-19
fix(TDI-45139): bigquery numeric types

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-45139

Fix Numeric type handling.

PR #527, opened 2020-12-28
fix(TDI-45139): clean pr commits

mbasiuk-talend:
Re-creating https://github.com/Talend/connectors-se/pull/491.


dmytrogrygorenko (approved):
LGTM

PR #530, opened 2020-12-31
fix(TDI-45257):[inventory] - Salesforce "Picklist" fields are not available when creating a dataset

jzhao-talend:
For Jira: https://jira.talendforge.org/browse/TDI-45257


pyzhou-talend (approved):
LGTM

PR #531, opened 2021-01-05
feat(TDI-45206): implement guess schema action

undx:
https://jira.talendforge.org/browse/TDI-45206

### DIFF
--- a/jdbc/src/main/java/org/talend/components/jdbc/input/ChangeDataCaptureInputEmitter.java
+++ b/jdbc/src/main/java/org/talend/components/jdbc/input/ChangeDataCaptureInputEmitter.java
@@ -12,28 +12,32 @@
  */
 package org.talend.components.jdbc.input;
 
-import lombok.extern.slf4j.Slf4j;
+import static org.talend.components.jdbc.ErrorFactory.toIllegalStateException;
+import static org.talend.sdk.component.api.record.Schema.Type.RECORD;
+
+import java.io.Serializable;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.stream.IntStream;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+
 import org.talend.components.jdbc.configuration.InputCaptureDataChangeConfig;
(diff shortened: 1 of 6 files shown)


jdbc/src/main/java/org/talend/components/jdbc/input/ChangeDataCaptureInputEmitter.java
```diff
@@ -239,126 +245,6 @@ private String getRowAsString() throws Exception {
...
-            switch (sqlType) {
-            case java.sql.Types.SMALLINT:
-            case java.sql.Types.TINYINT:
-            case java.sql.Types.INTEGER:
-                if (value != null) {
-                    if (javaType.equals(Integer.class.getName())) {
-                        builder.withInt(entryBuilder.withType(INT).build(), (Integer) value);
-                    } else {
-                        builder.withLong(entryBuilder.withType(LONG).build(), (Long) value);
-                    }
```
ozhelezniak-talend, line 311: Just a small notice.
This part didn't support the Short time before.
I think a new implementation better, caz here we may have a class cast exception.

jdbc/src/main/java/org/talend/components/jdbc/input/ChangeDataCaptureInputEmitter.java
```diff
@@ -239,126 +245,6 @@ private String getRowAsString() throws Exception {
...
-                    .withNullable(metaData.isNullable(columnIndex) != columnNoNulls);
-            switch (sqlType) {
-            case java.sql.Types.SMALLINT:
-            case java.sql.Types.TINYINT:
-            case java.sql.Types.INTEGER:
-                if (javaType.equals(Integer.class.getName())) {
-                    builder.withEntry(entryBuilder.withType(INT).build());
-                } else {
-                    builder.withEntry(entryBuilder.withType(LONG).build());
-                }
```
ozhelezniak-talend, line 257: the same about previously unsupported Short type.
Instead of it we will have: 
`if (javaType.equals(Integer.class.getName()) || Short.class.getName().equals(javaType)) {
                    builder.withEntry(entryBuilder.withType(INT).build());
                } else {
                    builder.withEntry(entryBuilder.withType(LONG).build());
                }`

ozhelezniak-talend (approved):
LGTM

PR #541, opened 2021-01-25
fix(TDI-45412) : null excel cells are not set to prevent NPE

ypiel-talend:
https://jira.talendforge.org/browse/TDI-45412

**New behavior:**
If there is a null value on a primitive type, we ignore the cell (default value ou let it blank).


OleksiiNimych (dismissed):
LGTM

OleksiiNimych (approved):
LGTM

OleksiiNimych (approved):
LGTM

OleksiiNimych (approved):
LGTM

PR #548, opened 2021-02-02
TDI-45585 : JUnit 4 to 5

clesaec:
https://jira.talendforge.org/browse/TDI-45585

remove all element of JUnit4 and replace it with JUnit5
In practice, it only concern assert statement, never @Test 
No impact on tests


chmyga (approved):
LGTM

PR #549, opened 2021-02-02
feat(TDI-45335): Bypass rate limit on docker hub

undx:
https://jira.talendforge.org/browse/TDI-45335


ozhelezniak-talend (approved):
LGTM

PR #553, opened 2021-02-09
Clesaec/tdi 45389 clean pom

clesaec:
https://jira.talendforge.org/browse/TDI-45389

First step to clean pom on connectors-se module

PR #556, opened 2021-02-16
Dsylaiev/tdi 42576 refactor azure and adls new

dmytro-sylaiev:


ovladyka (dismissed):

clesaec:
- Good idea to group azure-blob & azure-dls-gen2 on same sub module. Just wait my [PR](https://github.com/Talend/connectors-se/pull/553) as for the moment, jenkins doen't show unit test result on [azure sub module](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/dsylaiev%252FTDI-42576_refactor_azure_and_adls_new/16/testReport/) (575 tests, vs [848](https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/clesaec%252FTDI-45389_cleanPOM/42/testReport/) just by changing poms)

- azure/azure-common module contains only classes that are either specific to azure-blob or to azure-dls-gen2 ... So, there are no really common thing on this module ?

- On connectors-common module (any reason to rename common); classes converters are duplicated of common-stream-io (org.talend.components.common.converters.AvroConverters & org.talend.components.common.stream.input.avro.AvroToRecord for example). 
So, 2 possibility : 
1 - replace usage of AvroConverters with AvroToRecord/RecordToAvro; or use directly AvroOutput 
2 - or, if too complicate (need migration task ...), move "common.converter" module under "azure/azure-common" one (and then, this module will contains real common classes).

Same for common/format (either use common-stream-io/common-io; either move this azure specific format to azure-common)

rlecomte-talend (approved):
LGTM

rlecomte-talend (approved):
LGTM

PR #557, opened 2021-02-18
feat(TDI-45586): add support for column selection

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-45586

Make columns property. Select needed columns and use them.


dmytrogrygorenko (approved):
LGTM

PR #567, opened 2021-03-04
fix(TDI-45712): fix NPE, add test

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-45712

In the complete method, there was a call performed on the null batch list.
Add an additional check for a non-null object.


dmytrogrygorenko (dismissed):
LGTM

dmytrogrygorenko (approved):

PR #582, opened 2021-04-01
TDI-45729 column mismatch

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-45729

Fix column mismatch in copy into query.


chmyga (approved):
LGTM

PR #611, opened 2021-05-19
Rlecomte/tdi 46042 azure dls gen2 large files size

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-46042

Changed blob size from integer to long as there was an issue with blobs > 2GB


dmytro-sylaiev (approved):
LGTM

PR #619, opened 2021-06-07
fix(TDI-46185): MySQL Connection issues due to TLS 1.2 after java upgrade to 1.8.0_292

wwang-talend:


wwang-talend:
junit failed as network, will trigger it again.

wyuan (approved):

PR #626, opened 2021-06-10
TDI-46281 : add iterator filter

clesaec:
https://jira.talendforge.org/browse/TDI-46281

Add simple common facilities to filter values on iterator.

### DIFF
--- a/common/src/test/java/org/talend/components/common/collections/IteratorComposerTest.java
+++ b/common/src/test/java/org/talend/components/common/collections/IteratorComposerTest.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.function.Supplier;
+import java.util.stream.StreamSupport;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


common/src/test/java/org/talend/components/common/collections/IteratorComposerTest.java
```diff
@@ -54,6 +55,23 @@ public void flatmap() {
...
+        final Iterator<Integer> iterator = IteratorComposer.of(strings.iterator()) //
+                .filter((String s) -> s.length() > 2) //
+                .map(String::length) //
+                .build();//
+        int nbe = 0;
+        while (iterator.hasNext()) {
+            nbe++;
+            final Integer next = iterator.next();
+            Assertions.assertTrue(next.intValue() > 2);
+            Assertions.assertTrue(nbe <= 3);
```
chmyga, line 70: this check is redundant as the number is checked after the cycle.

chmyga (approved):
Just one small comment which isn't a problem. LGTM.

common/src/test/java/org/talend/components/common/collections/IteratorComposerTest.java
clesaec: it's to avoid infinite loop in case of bug in class code.

PR #630, opened 2021-06-18
TDI-46313 : remove schema impl ref in unit test

clesaec:
https://jira.talendforge.org/browse/TDI-46313

=> code must not reference SchemaImpl nor EntryImpl nor RecordImpl in unit test


rlecomte-talend (approved):
LGTM

PR #646, opened 2021-07-12
fix(TDI-45850): align "joda-time" version for connectors-se

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-45850
Align version of "joda-time" to use version defined in parent pom.


mbasiuk-talend (dismissed):
LGTM

mbasiuk-talend (approved):

PR #649, opened 2021-07-16
fix(TDI-46403): couchbase primary index issue

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-46403
Add checkbox to control Primary Index creation.
Default is "false", migration task included.


chmyga (approved):
LGTM

PR #651, opened 2021-07-20
fix(TDI-46442): rename "Analytics" option.

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-46442
Rename "Analytics" option to "N1QL for Analytics".


chmyga (approved):
LGTM

PR #667, opened 2021-08-11
fix(TDI-46567): initialize List to avoid NPE.

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-46567
Need to initialize List to avoid NPE during job run.


mbasiuk-talend (approved):
LGTM

PR #674, opened 2021-09-01
fix(TDI-46614): update "commons-compress" to 1.21

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-46614

CVE - update "commons-compress" to 1.21


mbasiuk-talend (approved):
LGTM

PR #676, opened 2021-09-03
TDI-46681 : Add log on Record IO Repo

clesaec:
https://jira.talendforge.org/browse/TDI-46681 : trace on loaded format & error when try to get null.


chmyga (approved):
LGTM

PR #681, opened 2021-09-13
fix(TDI-46715): Check Jdbc input query is one single query

arnaudfnr:
As explained in https://jira.talendforge.org/browse/TDI-46715, 
When a user create a dataset using jdbc connection and an input query,
he can chain queries in a one-liner using the `;` separator.
Doing so, the read-only check is bypassed and a user can performed non readonly operations.

This fix is a quick/dirty fix, there are certainly more uncovered use cases and we can't find them all using regexes.
However the new SQL compiler of Processing Runtime will be using a SQL parser (Apache Calcite), therefore a better solution might be found then.

### DIFF
--- a/jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
+++ b/jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
@@ -73,7 +73,10 @@ public abstract class AbstractInputEmitter implements Serializable {
         if (query == null || query.trim().isEmpty()) {
             throw new IllegalArgumentException(i18n.errorEmptyQuery());
         }
-        if (jdbcDriversService.isNotReadOnlySQLQuery(query)) {
+        if (jdbcDriversService.isNotReadOnlySQLQuery(query) || query.contains(";")) {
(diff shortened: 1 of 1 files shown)


Gby56 (approved):
LGTM

jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
```diff
@@ -73,7 +73,10 @@ public void init() {
         if (query == null || query.trim().isEmpty()) {
             throw new IllegalArgumentException(i18n.errorEmptyQuery());
         }
-        if (jdbcDriversService.isNotReadOnlySQLQuery(query)) {
+        if (jdbcDriversService.isNotReadOnlySQLQuery(query) || query.contains(";")) {
```
ypiel-talend, line 76: Too restrictive since the following acceptable query:
select * from a_field='some;content'
will be refused.

ypiel-talend (changes requested):
There are some comments in the Jira.

Gby56:
Apache Calcite will likely be required to properly interpret the input and detect 1 or more queries then

PR #686, opened 2021-09-16
Rlecomte/tdi 46360 enforce https

clesaec:
https://jira.talendforge.org/browse/TDI-46360
Enforce secure protocol (depends on env)


common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -12,74 +12,150 @@
...
+    }
+
+    @RequiredArgsConstructor
+    public enum AccessControls {
+        LOCAL(
+                (final URL url, final InetAddress inetAddress) -> !inetAddress.isSiteLocalAddress()
+                        && !inetAddress.isLoopbackAddress()
+                        && ADDITIONAL_LOCAL_HOSTS.stream()
+                                .noneMatch(h -> url.getHost().contains(h)),
+                (Environment env) -> env.getValue("CONNECTORS_ENABLE_LOCAL_NETWORK_ACCESS",
```
chmyga, line 88: Wouldn't it be better to move constant strings to constant fields?

chmyga (approved):
Just a small comment, otherwise LGTM

chmyga (approved):
LGTM

coheigea:
@clesaec Are you planning to enforce other schemes such as pulsar/hdfs as in the JIRA? Did you check all the connectors to make sure any of the others should enforce a protocol?

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
```diff
@@ -12,74 +12,172 @@
...
+                        Constants.JVM_ENABLE_LOCAL_NETWORK_ACCESS, "true"),
+                "[local]"),
+
+        MULTICAST(
+                (final URL url, final InetAddress inetAddress) -> !(inetAddress.isMulticastAddress()),
+                (Environment env) -> env.getValue(Constants.ENV_ENABLE_MULTICAST_NETWORK_ACCESS,
+                        Constants.JVM_ENABLE_MULTICAST_NETWORK_ACCESS, "true"),
+                "[multicast]"),
+
+        NON_SECURED(
```
coheigea, line 120: I'm not sure about enforcing https via a system property here. Shouldn't we allow the customer to specify a http:// URL in the connector if they have it configured? The task was more to prevent someone using a protocol they shouldn't for a given connector.

clesaec:
For the others connectors, pulsars, hdfs, that are not in connectors-se, JIRA https://jira.talendforge.org/browse/TDI-46730 and https://jira.talendforge.org/browse/TDI-46732 are written.

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
clesaec: for REST connector, we plan to use system property to see if we enforce https, by using this function (same for cosmosDB and salesforces).
For pulsar, same system property will enforce customer to use "pulsar+ssl://" protocol.

coheigea: The problem is if the system property is not enforced, someone could still try to use a protocol that's not supported for the connector (e.g. tcp://)

salesforce/src/main/java/org/talend/components/salesforce/service/SalesforceService.java
```diff
@@ -175,14 +164,22 @@ protected String getEndpoint(final BasicDataStore datastore, final LocalConfigur
...
-            return endpoint;
+            realEndPoint = endpoint;
         } else {
-            return URL;
+            realEndPoint = URL;
+        }
+        if (!ValidateSites.isValidSite(realEndPoint)) {
+            log.warn("Not valid base {}", realEndPoint);
+            final String errorMessage = ValidateSites.buildErrorMessage(messages::notValidAddress, realEndPoint);
+            throw new ComponentException(ComponentException.ErrorOrigin.USER, errorMessage);
```
jzhao-talend, line 180: after this change, seems : 
https://github.com/Talend/connectors-se/blob/272d644b344864b2d289767a50c1f12b6d710ed8/salesforce/src/main/java/org/talend/components/salesforce/service/UiActionService.java#L62-L74
the validateBasicConnection code also need to adjust follow this change.

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
coheigea: @clesaec Will my comment be addressed?

clesaec: If protocol is not supported by the connector where it is used, then connector itself will generate an exception ? (as i you try to put http on pulsar instead of pulsar protocol or specific control for ftp connector); 
or may i didn't understand comment about tcp ?

salesforce/src/main/java/org/talend/components/salesforce/service/SalesforceService.java
clesaec: Changed

common/src/main/java/org/talend/components/common/service/http/ValidateSites.java
coheigea: What I meant was that the connector could support a protocol that we don't want to support in Talend Cloud. For example for HDFS you can specify "file:///" instead of "hdfs".

clesaec: Inside TCK connector, we can't know if we are running in "Cloud" (via pipeline designer or cloud studio) or in Local (Studio local) where file protocol can be allowed.
For hdfs connector, i think it's more the role of [fileio/hdfs](https://github.com/Talend/cloud-components/tree/master/fileio/src/main/java/org/talend/components/fileio/hdfs) connector to control this specific protocol rather than a common module ?

coheigea: Yep I agree, can we enforce them instead in the individual connectors?

chmyga (approved):
LGTM

PR #692, opened 2021-10-04
fix(TDI-46745): CVE jsoup - update to 1.14.2

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-46745
Fix for CVE issue - align version of "jsoup" with parent pom.


mbasiuk-talend (approved):
LGTM

PR #694, opened 2021-10-04
Dsylaiev/tdi 46822 revert azure refactoring

dmytro-sylaiev:
Need to revert since several regressions appear in TDI-42576


ypiel-talend (approved):
LGTM

PR #710, opened 2021-10-21
fix(TDI-46843): Change the error message for parquet azureblob ad auth

dmytro-sylaiev:


clesaec (dismissed):
:+1:

azure/azureblob/src/main/java/org/talend/components/azure/runtime/input/BlobFileReader.java
```diff
@@ -81,10 +82,10 @@ public CloudBlobContainer checkBlobContainer(AzureBlobDataset config, CloudBlobC
 
         try {
             if (!container.exists()) {
-                throw new BlobRuntimeException(messageService.containerNotExist());
+                throw new ComponentException(messageService.containerNotExist());
```
clesaec, line 85: Add the name of container that is not exists in message : 
`messageService.containerNotExist(config.getContainerName())`
With containerNotExist(String name) method in interface and a {0} symbol in Messages.properties file.

clesaec (changes requested):
Just one remark

azure/azureblob/src/main/java/org/talend/components/azure/runtime/input/BlobFileReader.java
dmytro-sylaiev: it doesn't work for some reason

dmytro-sylaiev: the error message on the PD still contains "{0}" instead of a parameter

clesaec: Strange, it works on other connector : [jdbc for example](https://github.com/Talend/connectors-se/blob/master/jdbc/src/main/resources/org/talend/components/jdbc/service/Messages.properties#L4)

clesaec (dismissed):
:+1:

clesaec (approved):
:+1:

PR #713, opened 2021-11-01
fix(TDI-45416): Fix excel html for custom encoding

dmytro-sylaiev:
https://jira.talendforge.org/browse/TDI-45416


azure/azurecommon/src/main/java/org/talend/components/common/formats/excel/ExcelFormatOptions.java
```diff
@@ -75,4 +78,22 @@
     @Documentation("")
     @Min(0)
     private int footer = 1;
+
+    public String effectiveHTMLFileEncoding() {
```
clesaec, line 82: Throw directly a ComponentException, 
To do that with correct msg translation, change here 
`String effectiveHTMLFileEncoding(final Function<String, String> errorMsgBuilder)`

So, you call it with  `effectiveHTMLFileEncoding(this.getMessageService()::encodingMessage);` in ExcelHTMLBlobFileReader class.
That will make message to the final user easier to understand.

azure/azurecommon/src/main/java/org/talend/components/common/formats/excel/ExcelFormatOptions.java
```diff
@@ -75,4 +78,21 @@
...
+        if (excelFormat != ExcelFormat.HTML) {
+            throw new IllegalStateException("Only HTML format supports encoding configuration"); // should not be here
+        }
+        if (Encoding.OTHER == getEncoding()) {
+            try {
+                Charset.forName(customEncoding);
+                return getCustomEncoding();
+            } catch (Exception e) {
+                log.error("[effectiveFileEncoding] {}", e.getMessage());
+                throw new ComponentException(errorMsgBuilder.apply(errorMsgBuilder.apply(getCustomEncoding())));
```
clesaec, line 92: Why call twice errorMsgBuilder.apply ?

clesaec (approved):
:+1:

PR #738, opened 2021-12-09
fix(TDI-47231): Add tabulation migration for adls-gen2

dmytro-sylaiev:
https://jira.talendforge.org/browse/TDI-47231


dmytrogrygorenko (approved):
LGTM

PR #749, opened 2021-12-23
feat(TDI-41801): extract common codes from mongodb

vyu-talend:
https://jira.talendforge.org/browse/TDI-41801


mongo/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+        <version>1.30.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>mongo</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Connectors SE :: Mongo</name>
+
+    <modules>
+        <module>mongo-common</module>
```
jzhao-talend, line 17: need check together about the module name, IMO, mongo is not an official name
I prefer:
mongodb
  -mongodb-common
  -mongodb-connector

this would change the artificid, need to discuss it together

mongo/mongo-common/src/main/java/org/talend/components/mongo/Address.java
```diff
@@ -10,7 +10,7 @@
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
-package org.talend.components.mongodb;
+package org.talend.components.mongo;
```
jzhao-talend, line 13: i prefer to keep mongodb, need to check together with the module name

mongo/pom.xml
vyu-talend: If no technical problem, I think we can get QA to start testing. How do you think about it? @jzhao-talend

mongo/mongo-common/src/main/java/org/talend/components/mongo/source/MongoCommonSourceConfiguration.java
```diff
@@ -10,16 +10,17 @@
...
-import org.talend.components.mongodb.datastore.MongoDBDataStore;
+import org.talend.components.mongo.dataset.MongoCommonDataSet;
 
 import java.io.Serializable;
 
-public interface BaseSourceConfiguration extends Serializable {
+public interface MongoCommonSourceConfiguration extends Serializable {
 
-    BaseDataSet getDataset();
+    void setDataset(MongoCommonDataSet dataset);
```
jzhao-talend, line 21: seems no need add this method in the interface.

mongo/mongodb/src/main/java/org/talend/components/mongodb/sink/MongoDBProcessor.java
```diff
@@ -63,22 +59,18 @@
...
 
-    private transient RecordToJson recordToJson;
-
-    private transient RecordToDocument recordToDocument;
-
     public MongoDBProcessor(@Option("configuration") final MongoDBSinkConfiguration configuration,
             final MongoDBService service,
             final I18nMessage i18n) {
-        this.configuration = configuration;
+        super.configuration = this.configuration = configuration;
```
jzhao-talend, line 65: you configured a protected field in the father class.
i think you can use it directly in the child class.

vyu-talend: yes, but it still needs to define a MongoDBSinkConfiguration object, and force cast the MongoCommonOutputConfiguration to MongoDBSinkConfiguration. So I think it would be a little more concise in that way.

mongo/mongo-common/src/main/java/org/talend/components/mongo/source/MongoCommonSourceConfiguration.java
vyu-talend: Yes, there is no related code in the runtime. But it is needed in the junit test, please check [here](https://github.com/Talend/connectors-se/pull/749/files#diff-62e588e2a34e45f0e8c7ec0d4916ec96904c3ba0a15c65408561119971ea0800R230)

mongo/mongodb/src/test/java/org/talend/components/mongodb/MongoDBTest.java
```diff
@@ -236,8 +225,8 @@ void testBasic() {
         Assertions.assertEquals("123.123", record.get(Object.class, "number"));
     }
 
-    private List<Record> getRecords(MongoDBReadDataSet dataset) {
-        MongoDBQuerySourceConfiguration config = new MongoDBQuerySourceConfiguration();
+    private List<Record> getRecords(MongoCommonDataSet dataset) {
```
jzhao-talend, line 228: i think you can keep the code before you change it.
i don't see this method called with other sub-class.
all the code which called this method passed `MongoDBReadDataSet`

you also have another method for `MongoDBReadAndWriteDataSet ` https://github.com/Talend/connectors-se/pull/749/files#diff-62e588e2a34e45f0e8c7ec0d4916ec96904c3ba0a15c65408561119971ea0800R236

mongo/mongo-common/src/main/java/org/talend/components/mongo/source/MongoCommonSourceConfiguration.java
jzhao-talend: https://github.com/Talend/connectors-se/pull/749/files#r782828848

mongo/pom.xml
jzhao-talend: checked with Pierre, we prefer to:
mongodb (artifactId change to `mongodb-parent`)
-mongodb-common
-mongodb

pyzhou-talend (approved):
LGTM

PR #756, opened 2022-01-03
fix(TDI-47329): Azure-dls-gen2 - TU fail after support the latest API version

ypiel-talend:
https://jira.talendforge.org/browse/TDI-47329


dmytro-sylaiev (approved):
LGTM

PR #760, opened 2022-01-04
feat(TDI-46093): Add connection component for Tck ADLS gen2

ypiel-talend:
https://jira.talendforge.org/browse/TDI-46093


Tanya5091 (approved):
LGTM

PR #773, opened 2022-01-12
fix(TDI-47371):Snowflake upsert delete cannot work with special char

pyzhou-talend:
…cter
https://jira.talendforge.org/browse/TDI-47371


jzhao-talend (approved):
LGTM

PR #778, opened 2022-01-20
Wwang talend/tdi 47408

wwang-talend:


jzhao-talend:
TDI-47408 has been in final check, so better close this jira, as we don't need this currently.
if need this change in the feature, please create a new PR with a new jira.

PR #789, opened 2022-02-08
fix(TDI-47485): update "protobuf-java" to 3.19.2

dmytrogrygorenko:
https://jira.talendforge.org/browse/TDI-47485
Update "protobuf-java" library to 3.19.2 to fix CVE.


azure/pom.xml
```diff
@@ -37,11 +37,6 @@
 
   <dependencyManagement>
     <dependencies>
-      <dependency>
-        <groupId>com.google.protobuf</groupId>
-        <artifactId>protobuf-java</artifactId>
-        <version>3.4.0</version>
-      </dependency>
```
clesaec, line 44: Why not putting this managed dependency directly in connectors-se [root pom](https://github.com/Talend/connectors-se/blob/master/pom.xml#L163) ?

dmytrogrygorenko: That's what I did, please look below in this PR.

clesaec (dismissed):
:+1:

clesaec (approved):

PR #794, opened 2022-02-15
fix(TDI-47349): do a try to exclude it

wwang-talend:


wwang-talend:
https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/wwang-talend%252FTDI-47349/

PR #803, opened 2022-02-22
chore(TDI-47520): comment marker

ozhelezniak-talend:
https://jira.talendforge.org/browse/TDI-47520

Add a new option in the common part (common-stream-io). 
it allows specific comment marker (the lines that start from this character will be recognized as comments)
The option works only on Write action when the user wants to write a header. 
For Read operation, the behavior wasn't changed. The comments are disabled.

### DIFF
--- a/common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CSVConfiguration.java
+++ b/common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CSVConfiguration.java
@@ -21,8 +21,15 @@ import org.talend.sdk.component.api.meta.Documentation;
 import lombok.Data;
 
 @Data
-@GridLayout({ @GridLayout.Row("lineConfiguration"), @GridLayout.Row({ "fieldSeparator" }) })
-@GridLayout(names = GridLayout.FormType.ADVANCED, value = { @GridLayout.Row({ "escape", "quotedValue" }) })
+@GridLayout({ //
+        @GridLayout.Row({ "lineConfiguration" }), //
+        @GridLayout.Row({ "fieldSeparator" })
+})
+@GridLayout(names = GridLayout.FormType.ADVANCED,
+        value = { //
+                @GridLayout.Row({ "escape", "quotedValue" }),
+                @GridLayout.Row({ "commentMarker" })
+        })
 public class CSVConfiguration implements ContentFormat {
 
     private static final long serialVersionUID = -6803208558417743486L;
@@ -40,6 +47,10 @@ public class CSVConfiguration implements ContentFormat {
     private Character escape = '\\';
 
     @Option
+    @Documentation("Comment marker.")
+    private CommentMarker commentMarker = new CommentMarker();
+
+    @Option
     @Documentation("Text enclosure character.")
     private Character quotedValue = '"';
 
@@ -48,5 +59,13 @@ public class CSVConfiguration implements ContentFormat {
             return ';';
         }
         return this.fieldSeparator.findFieldSeparator();
+    }
+
+    public Character findCommentMarker() {
+        if (commentMarker == null) {
+            return ' '; // the default value that was used
+        }
+
+        return commentMarker.findCommentMarker();
     }
 }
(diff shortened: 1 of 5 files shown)


common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CommentMarker.java
```diff
@@ -0,0 +1,64 @@
...
+        COMMA(','),
+        SPACE(' '),
+        HASH('#'),
+        OTHER((char) 0);
+
+        private final char marker;
+    }
+
+    @Option
+    @Documentation("Comment marker type.")
```
clesaec, line 49: Add annotation
@DefaultValue("SPACE")
(or may be better default with NONE to keep current behavior ?)

common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CSVConfiguration.java
```diff
@@ -39,6 +46,10 @@
     @Documentation("Escape character.")
     private Character escape = '\\';
 
+    @Option
+    @Documentation("Comment marker.")
+    private CommentMarker commentMarker = new CommentMarker();
```
clesaec, line 51: This will be visible for all connector using CSV, in input & output. So, it could be better to use it directly in CSVHelper. So reader will (should) ignore commented lines.

common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CommentMarker.java
```diff
@@ -0,0 +1,64 @@
...
+})
+public class CommentMarker implements Serializable {
+
+    private static final long serialVersionUID = 406534099976881354L;
+
+    @Getter
+    @RequiredArgsConstructor
+    public enum Type {
+
+        SEMICOLON(';'),
```
clesaec, line 39: May add a NONE to keep behavior of existing connector ?

clesaec (commented):
Some questions.
May be we can have a try with a more recent version of commons-csv, [1.9.0 is the last](https://mvnrepository.com/artifact/org.apache.commons/commons-csv/1.9.0).

common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CommentMarker.java
ozhelezniak-talend: In the annotation description, it says that we can use @DefaultValue only for primitive types.
Will it work with enumbs?

ozhelezniak-talend: Maybe NONE is a good idea. 
And add the exception\validation that when the user uses header they must specify comment marker. 
It's a change. It's breaking change but it's less harmful in the future perspective. (IMO)

common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CSVConfiguration.java
ozhelezniak-talend: Yeah, thank you, I thought about that. It's one of the things that I wanted to discuss. 

Looks like right now CSVReader doesn't know about comments. If I understood it correct the default behavior is setCommentMarker(null) == (disabled).
And it'll be **a breaking change** for the connectors that used the stream-csv if I add support of comment marker in CSVReader. 
--- 
Only in the complex with NONE and let NONE take all breaking change responsibility

common-stream-io/common-io/src/main/java/org/talend/components/common/stream/format/csv/CommentMarker.java
ozhelezniak-talend: Yes, maybe. But it won't completely cover the default behavior. 
The behavior of the existing connector is to use SPACE as the default comment marker when we use headers too.

ozhelezniak-talend:
Okay. Now I'm considering adding NONE state and adding more tests for reading part.

PR #808, opened 2022-03-15
fix(TDI-47538): tAzureAdlsGen2Output component converts BigDecimal to String

jzhao-talend:
for: https://jira.talendforge.org/browse/TDI-47538

tAzureAdlsGen2Output write and read bigdecimal type in parquet type

### DIFF
--- a/common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
+++ b/common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
@@ -30,5 +30,7 @@ public class Constants {
 
     public static final String STUDIO_PRECISION = "talend.studio.precision";
 
+    public static final String BIGDECIMAL = "id_BigDecimal";
+
     public static final String STUDIO_TYPE = "talend.studio.type";
 }
--- a/common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
+++ b/common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
@@ -10,17 +10,25 @@
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
-package org.talend.components.common.stream.input.avro;
+package org.talend.components.common.stream;
 
 public class Constants {
 
     public static final String AVRO_LOGICAL_TYPE = "logicalType";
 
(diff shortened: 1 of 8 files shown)


common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public static final String AVRO_LOGICAL_TYPE_TIME_MILLIS = "time-millis";
+
+    public static final String AVRO_LOGICAL_TYPE_TIMESTAMP_MILLIS = "timestamp-millis";
+
+    public static final String AVRO_LOGICAL_TYPE_DECIMAL = "decimal";
+
+    public static final String ERROR_UNDEFINED_TYPE = "Undefined type %s.";
+
+    public static final String STUDIO_LENGTH = "talend.studio.length";
```
clesaec, line 33: Why named this constant with "Studio" ? As it is used for Avro & Parquet for Pipeline deigner & studio.
Should be better with
`public static final String DECIMAL_LENGTH = "talend.decimal.length";`
for example ?

clesaec (approved):
Suggest one small change about constant Names.

common-stream-io/stream-avro/src/main/java/org/talend/components/common/stream/Constants.java
jzhao-talend: https://github.com/Talend/component-runtime/pull/551/files#diff-b2dc0f214b21ca9bf02841caeb3774827c739f648ee10bd74386fac92137366eR21-R27
should be the same with component-runtime

jzhao-talend: so component-runtime-di also needs to adjust to a generic name. event it only for the studio.
during avro convert, we need to get the decimal properties from the schema generated by component-runtime-di

clesaec: ok, i see

clesaec:
Just to be sure, could you add a test 'record to Avro' with an array of BigDecimal element

jzhao-talend:
> 

what about the talend type are you talking about?
List or Object?

clesaec:
I mean a org.apache.avro.Schema.Type.ARRAY of org.apache.avro.Schema.Type.BYTES that have Constants.AVRO_LOGICAL_TYPE_DECIMAL logical types.

clesaec (approved):

clesaec (approved):

PR #815, opened 2022-03-23
fix(TDI-47708): fix line terminators in messages properties files

davroussel:
Following https://jira.talendforge.org/browse/TDI-47708:
* add an empty line at the end of file
* change EOL terminator from CRLF to LF
* remove empty files


ypiel-talend (dismissed):
LGTM

quilicicf (approved):

PR #818, opened 2022-03-28
fix(TDI-46715): JDBC connector non readonly queries error

dochkas-talend:
Added Apache Calcite dependency and added nodes count validation

### DIFF
--- a/jdbc/src/test/java/org/talend/components/jdbc/suite/JDBCBaseContainerTest.java
+++ b/jdbc/src/test/java/org/talend/components/jdbc/suite/JDBCBaseContainerTest.java
@@ -532,6 +532,27 @@ public abstract class JDBCBaseContainerTest {
         }
 
         @Test
+        @DisplayName("Query -  non authorized query (multiple queries)")
+        void unauthorizedMultipleQueries() {
+            final SqlQueryDataset dataset = new SqlQueryDataset();
+            dataset.setConnection(newConnection());
+            dataset.setSqlQuery("SELECT * FROM my_table; SELECT * FROM my_table;");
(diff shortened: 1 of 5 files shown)


jdbc/src/test/java/org/talend/components/jdbc/suite/JDBCBaseContainerTest.java
```diff
@@ -531,6 +531,27 @@ void unauthorizedInsertQuery() {
                             .run());
         }
 
+        @Test
+        @DisplayName("Query -  non authorized query (multiple queries)")
+        void unauthorizedMultipleQueries() {
+            final SqlQueryDataset dataset = new SqlQueryDataset();
+            dataset.setConnection(newConnection());
+            dataset.setSqlQuery("SELECT * FROM my_table; SELECT * FROM my_table;");
```
ypiel-talend, line 539: Should not we have a multi queries test for each supported database  to check specificities like character enclosure " or '  ?
I think we should use: https://github.com/Talend/connectors-se/blob/1b4c89c1f253bb46d49d30849b4074fe25c565e4/jdbc/src/main/java/org/talend/components/jdbc/output/platforms/Platform.java#L147
To generate 2 queries joinned with ';'.

(This how is build TableNameDataset query: https://github.com/Talend/connectors-se/blob/1b4c89c1f253bb46d49d30849b4074fe25c565e4/jdbc/src/main/java/org/talend/components/jdbc/dataset/TableNameDataset.java#L69 )

ypiel-talend (commented):
One comment

ypiel-talend (approved):
LGTM

PR #831, opened 2022-04-21
fix(TDI-46961): Throw clear exception

dmytro-sylaiev:


azure/azureblob/src/main/java/org/talend/components/azure/runtime/input/ExcelHTMLBlobFileReader.java
```diff
@@ -15,6 +15,7 @@
 import java.io.InputStream;
 import java.net.URISyntaxException;
 import java.util.Iterator;
+import java.util.function.Function;
```
clesaec, line 18: Is it needed ? may pass optimize import.

clesaec (dismissed):
Approved.
Later, you may replace Supplier<String> with Function<String, String> to details cause of HTML non validity (empty file, non regular number of rows, and may be others reasons that are currently encapsulate in Runtime Exception at ExcelHTMLBlobFileReader, line 82-84 and that depends of Jsoup parser)

dmytrogrygorenko (approved):
LGTM

PR #832, opened 2022-04-22
fix(TDI-47426):fix docdb won't check conn on runtime issue.

vyu-talend:
https://jira.talendforge.org/browse/TDI-47426


pyzhou-talend (approved):
LGTM

jzhao-talend (approved):

PR #838, opened 2022-04-29
feat(TDI-47912): bump component-runtime to 1.44.3

undx:
https://jira.talendforge.org/browse/TDI-47912

**Related PR**
- https://github.com/Talend/connectors-ee/pull/625
- https://github.com/Talend/connectors-se/pull/838
- https://github.com/Talend/cloud-components/pull/991

### DIFF
--- a/common-stream-io/stream-json/src/test/java/org/talend/components/common/stream/input/json/JsonToSchemaTest.java
+++ b/common-stream-io/stream-json/src/test/java/org/talend/components/common/stream/input/json/JsonToSchemaTest.java
@@ -12,6 +12,8 @@
  */
 package org.talend.components.common.stream.input.json;
 
+import static org.apache.avro.Schema.Type.NULL;
+
 import javax.json.Json;
 import javax.json.JsonNumber;
 import javax.json.JsonObject;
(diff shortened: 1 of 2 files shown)


common-stream-io/stream-json/src/test/java/org/talend/components/common/stream/input/json/JsonToSchemaTest.java
```diff
@@ -119,6 +124,15 @@ void inferBigSchema() throws IOException {
 
     }
 
+    public static org.apache.avro.Schema unwrapUnion(final org.apache.avro.Schema schema) {
```
ozhelezniak-talend, line 127: It's strange to see this change in regular bump of component-runtime for connectors-se
Is it related change?

undx: It's a fix from chris which was discussed in daily and accepted for putting it in the bump as it's tightly related to the bump.

ozhelezniak-talend (approved):
lgtm

PR #840, opened 2022-05-03
feat(TDI-47634): update talend i18n config file and remove locale version

davroussel:
* update talend i18n config file to include i18n resource in repository
* remove locales version in pom.xml


ypiel-talend (commented):
Should be review & merge only once thise one is merged: https://github.com/Talend/i18n-product/pull/162#pullrequestreview-989148975

ypiel-talend (approved):
LGTM

PR #843, opened 2022-05-04
chore(TDI-47931): set env var TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX

ypiel-talend:
https://jira.talendforge.org/browse/TDI-47931


Jenkinsfile
```diff
@@ -227,6 +227,7 @@ pipeline {
                         withCredentials([nexusCredentials
                                          , sonarCredentials]) {
                             sh """
+                                export TESTCONTAINERS_HUB_IMAGE_NAME_PREFIX='artifactory.datapwn.com/docker-io-remote'
```
undx, line 230: This env var is already present in Jenkinsfile 
https://github.com/Talend/connectors-se/blob/b7a67814063eea04a2e6cd11bee858a43c75f10f/Jenkinsfile#L93

undx: https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
> An environment directive used in the top-level pipeline block will apply to all steps within the Pipeline.

PR #857, opened 2022-05-18
fix(TDI-47975): Fix missing class for Delta adls

dmytro-sylaiev:
https://jira.talendforge.org/browse/TDI-47975


dmytrogrygorenko (approved):
LGTM

PR #870, opened 2022-06-08
i18n(TDI-47297): correct errorFileNotExist properties

jzhao-talend:
https://jira.talendforge.org/browse/TDI-47297


pyzhou-talend (approved):
LGTM

PR #877, opened 2022-06-20
chore(TDI-47867) Increase CI build visibility

fdesiles:
https://jira.talendforge.org/browse/TDI-47867

Updated Jenkinsfile to:

1. Only send slack messages to #component-ci for Master and Maintenance builds.
2. Ping with "here" when a build is failing and the previous one was OK.


ypiel-talend (approved):
LGTM

PR #879, opened 2022-06-21
fix(TDI-48107):gson CVE

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-48107


pom.xml
```diff
@@ -555,6 +555,11 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
```
jzhao-talend, line 558: where is it used in the connectors?

PR #901, opened 2022-08-17
fix(TDI-48414): add a unit test for RawTextRecordReader.

ypiel-talend:
https://jira.talendforge.org/browse/TDI-48414


common-stream-io/stream-rawtext/src/test/java/org/talend/components/common/stream/input/rawtext/RawTextRecordReaderTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+            String expected = baos.toString("UTF-8");
+
+            if(success) {
+                Assertions.assertEquals(expected, content);
+            }
+            else {
+                Assertions.assertNotEquals(expected, content);
+            }
+        } catch (IOException e) {
```
olhava, line 60: Maybe need to narrow down the try block to contain only operations that can actually throw IOException?

olhava (approved):

common-stream-io/stream-rawtext/src/test/java/org/talend/components/common/stream/input/rawtext/RawTextRecordReaderTest.java
ypiel-talend: I have removed the try..catch bloc. No need for the junit test.

olhava (approved):

PR #902, opened 2022-08-18
fix(TDI-48426):CVE: org.postgresql:postgresql:42.2.25

jzhao-talend:
https://jira.talendforge.org/browse/TDI-48426


pyzhou-talend (approved):
LGTM

PR #907, opened 2022-08-31
fix(TDI-47482): keep jdbc column order in UI

ozhelezniak-talend:
https://jira.talendforge.org/browse/TDI-47482

Just changed from toSet to toList. Not sure that columns can duplicate that's why I didn't use the suggested LinkedHashSet. I think a simple ArrayList would be enough.


undx (approved):
lgtm

PR #910, opened 2022-09-06
fix(TDI-48470):OS Command Injection in Talend/components (master) CVE-2022-25168:hadoop-common:3.2.3

jzhao-talend:
https://jira.talendforge.org/browse/TDI-48470


pyzhou-talend (approved):
LGTM

PR #912, opened 2022-09-07
doc(TDI-47628) Add Jira link to TODO

fdesiles:
Update POM with link to Jira to fix TODO:
TODO uncomment it to enable its on CI once creds are added (https://jira.talendforge.org/browse/TDI-47628)


acatoire (approved):
LGTM

PR #920, opened 2022-09-14
fix(TDI-48549): Unable to run TCK connector in Studio if BigDecimal Type is used in schema

wwang-talend:


undx (dismissed):
lgtm

clesaec:
Good fix i think, but add some unit test on that decimal.

jzhao-talend (approved):
LGTM

PR #923, opened 2022-09-22
Feat[TCOMP-2265]:Upgrade beam to 2.37.0

yyin-talend:
https://jira.talendforge.org/browse/TCOMP-2265


undx (approved):

undx:
Will be bumped when bumping component-runtime.

PR #938, opened 2022-10-14
fix(TDI-48634): rich tck/JDBC schema

wwang-talend:


wwang-talend:
https://jenkins-connectors.datapwn.com/job/connectors%20se/job/build/job/wwang-talend%252FTDI-48634/

jdbc/src/main/java/org/talend/components/jdbc/service/SchemaProperty.java
```diff
@@ -0,0 +1,33 @@
...
+    String SCALE = "scale";
+
+    // it's primary key for database, maybe also exists for other data source
+    String IS_KEY = "is.key";
+
+    // TODO they are special for database? we need to make it to common here?
+    String IS_FOREIGN_KEY = "is.foreign.key";
+
+    // TODO not special for database now, maybe better not make it so common here?
+    String IS_UNIQUE = "is.unique";
```
undx, line 31: I suggest to use a `field` namespace prefix for those properties. ie:
- `field.type`
- `field.size`
- `field.scale`
- `field.primary-key`
- `field.foreign-key`
- `field.unique`

wwang-talend: Hi @undx , how about this?
```
 String ORIGIN_TYPE = "field.type";

    String SIZE = "field.size";

    String SCALE = "field.scale";

    // it's primary key for database, maybe also exists for other data source
    String IS_KEY = "field.primary-key";

    // TODO they are special for database? we need to make it to common here?
    String IS_FOREIGN_KEY = "field. foreign-key";

    // TODO not special for database now, maybe better not make it so common here?
    String IS_UNIQUE = "is.unique";
```

wwang-talend: 1. but how to know the difference of meaning between tck type(STRING) and the data source type(VARCHAR2), field.type is a good key?

2. if the key is too long, it cost disk/hardware to store it in the middle process or final target, is good?

jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
```diff
@@ -118,8 +113,9 @@ public Record next() {
                 return null;
             }
             final Record.Builder recordBuilder = recordBuilderFactory.newRecordBuilder(schema);
-            for (int index = 0; index < columnInfoList.length; index++) {
-                jdbcDriversService.addColumn(recordBuilder, columnInfoList[index], resultSet.getObject(index + 1));
+            final List<Schema.Entry> columns = Optional.ofNullable(schema.getEntries()).orElse(Collections.emptyList());
+            for (int index = 0; index < columns.size(); index++) {
+                jdbcDriversService.addColumn(recordBuilder, columns.get(index), resultSet.getObject(index + 1));
```
undx, line 118: Using streams you can chain these statements...
(Maybe not relevant here) You may use `getAllEntries()` which includes metadata fields and return a stream.

jdbc/src/main/java/org/talend/components/jdbc/service/SchemaProperty.java
undx: 1) maybe something like `backend.type`, `field.native.type`, `field.native`, `field.origin.type`. (like the _native_ ones...)
2) I agree with that and I also think that properties should have a comprehensive name.
Anyway, it's better than actual in Studio w/ a namespace (`talend.studio.X`).

jdbc/src/main/java/org/talend/components/jdbc/service/UIActionService.java
```diff
@@ -282,6 +282,7 @@ public ValidationResult validateSortKeys(final RedshiftSortStrategy sortStrategy
         return new ValidationResult(ValidationResult.Status.OK, "");
     }
 
+    // TODO this method never been used
```
undx, line 285: This may be used by the dataset app to resolve schema when no data coming from sample. It has been asked for this purpose (but not aware if it's in effect yet...)

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
```diff
@@ -305,146 +298,256 @@ public static String getSchema(Connection connection) throws SQLException {
...
+                        String columnName = tableMetadata.getString("COLUMN_NAME");
+
+                        int size = tableMetadata.getInt("COLUMN_SIZE");
+                        int scale = tableMetadata.getInt("DECIMAL_DIGITS");
+                        int sqlType = tableMetadata.getInt("DATA_TYPE");
+                        boolean nullable = DatabaseMetaData.columnNullable == tableMetadata.getInt("NULLABLE");
+
+                        boolean isKey = keys.contains(columnName);
+
+                        // String defaultValue = tableMetadata.getString("COLUMN_DEF");
```
undx, line 360: unused?

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
```diff
@@ -305,146 +298,256 @@ public static String getSchema(Connection connection) throws SQLException {
...
+            // but will fail for the code below for table metadata fetch as table name is expected : [Company] without
+            // the full name prefix [catalog.schema1.].
+            String tableName = TableNameDataset.class.cast(dataSet).getTableName();
+            try {
+                catalog = connection.getCatalog();
+                dbSchema = JdbcService.getSchema(connection);
+
+                final DatabaseMetaData databaseMetadata = connection.getMetaData();
+
+                final Set<String> keys = getPrimaryKeys(databaseMetadata, catalog, dbSchema, tableName);
```
undx, line 347: unique and foreign-key metadata needed?

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
```diff
@@ -305,146 +298,256 @@ public static String getSchema(Connection connection) throws SQLException {
...
+                .withProp(SchemaProperty.ORIGIN_TYPE, String.valueOf(column.columnTypeName));
+
+        if (column.isKey) {
+            entryBuilder.withProp(SchemaProperty.IS_KEY, "true");
+        }
+
+        setSize(entryBuilder, column.size);
+
+        // TODO foreign key
+        // TODO is unique
```
undx, line 439: oups, sorry! forget my previous comment ;-)

jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
```diff
@@ -96,16 +94,13 @@ public void init() {
...
-            columnInfoList = IntStream.rangeClosed(1, metaData.getColumnCount())
-                    .mapToObj(index -> jdbcDriversService.addField(schemaBuilder, metaData, index))
-                    .toArray(JdbcService.ColumnInfo[]::new);
-            schema = schemaBuilder.build();
+            schema = jdbcDriversService.createSchema(dataSet, connection, resultSet, recordBuilderFactory);
 
             log.debug("Input schema: {}", schema);
-            log.debug("SchemaRaw: {}",
-                    schema.getEntries().stream().map(Schema.Entry::getRawName).collect(Collectors.joining(",")));
+            if (log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
wwang-talend: it just want to avoid to execute the statement and a small safe fix:
```
schema.getEntries().stream().map(Schema.Entry::getRawName).collect(Collectors.joining(",")))
```

jdbc/src/main/java/org/talend/components/jdbc/input/AbstractInputEmitter.java
wwang-talend: if use java streams api here, it need to process exception again inside, then not clear. 

this is cloud input connector runtime, jdbc runtime generate the schema here and ```getAllEntries()``` return metadata like row_id, no need that here. And here is a refactor, keep the old logic.

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
wwang-talend: that's TODO, i ask this in slack tdi-48633_complete_tck_schemas channel

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
wwang-talend: will clean it

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
```diff
@@ -305,146 +298,256 @@ public static String getSchema(Connection connection) throws SQLException {
...
-            case java.sql.Types.DECIMAL:
-            case java.sql.Types.NUMERIC:
-            case java.sql.Types.VARCHAR:
-            case java.sql.Types.LONGVARCHAR:
-            case java.sql.Types.CHAR:
-            default:
-                builder.withEntry(entryBuilder.withType(STRING).build());
-                break;
+            } catch (Exception e) {
+                log.error("[catalog,db schema,table] {} {} {}.", catalog, dbSchema, tableName);
```
wwang-talend, line 370: here we ignore the exception on purpose, as user's account may not have the rights to access that metadata info!

jdbc/src/main/java/org/talend/components/jdbc/service/SchemaProperty.java
wwang-talend: as the props are the field prop, so why need to ```field``` prefix?
how about :
```
origin.type
size
scale
primary-key
foreign-key
unique
```

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
wwang-talend: TODO : as we use DatabaseMetadata api:
1. some database have catalog&db schema both
2. some database only have catalog
3. some database only have db schema(not sure).

though we use connection.getSchema to get the db schema, but it depend on database and its jdbc implement. Sometimes, possible, we may get a empty string, or null even that database support db schema or catalog. Image we fetch this on oracle database with lots of db schemas:

```
databaseMetadata.getColumns("mydatabase", "", "a table", xxx)
```
that may be a very slow action as it will search all db schemas, we need to be careful that.

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
wwang-talend: done

jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
wwang-talend: test on mysql and oracle, is ok

undx (approved):
lgtm

pyzhou-talend (approved):
LGTM

PR #943, opened 2022-10-21
chore(TDI-47931): Fix TestContainer-Ryuk timeout with Docker in Docker

fdesiles:
https://jira.talendforge.org/browse/TDI-47931 ([internal] TestContainer: Can not connect to Ryuk)

* Update CI to use Docker In Docker (DinD) instead of Docker Sock.
* Disable JDBC Oracle integration test which is not working with new DinD -> Created ticket for fix : https://jira.talendforge.org/browse/TDI-48743


ypiel-talend (approved):
LGTM

PR #957, opened 2022-11-16
feat(TDI-48191): integrate connectors repos with XTM_auto_deploy job

davroussel:
See https://jira.talendforge.org/browse/TDI-48191 for details


ypiel-talend (approved):
LGTM

PR #960, opened 2022-11-17
feat(TDI-48895): fix capture files

undx:
fix headers mismatch


google-storage/src/test/resources/talend/testing/http/org.talend.components.google.storage.service.GSServiceTest_healthCheck.json
```diff
@@ -2,7 +2,7 @@
   {
     "request":{
       "headers":{
-        "content-length":"1076",
+        "Content-Length":"1076",
```
ypiel-talend, line 5: Headers should be case insensitive no ?
https://www.rfc-editor.org/rfc/rfc2616.html#section-4.2
"Field names are case-insensitive."

ypiel-talend (commented):
One question

google-storage/src/test/resources/talend/testing/http/org.talend.components.google.storage.service.GSServiceTest_healthCheck.json
ozhelezniak-talend: I met a problem once with those files and it was related to the netty framework and our http-junit module, and they used case-sensitive approach. 
We had a bug in tests because of that. The test stuck at that time. 
Not sure, that it's 100% related to this case, but it was a problem.

ozhelezniak-talend (approved):
lgtm
if tests pass on Jenkins

google-storage/src/test/resources/talend/testing/http/org.talend.components.google.storage.service.GSServiceTest_healthCheck.json
undx: Yes, headers _should_ be insensitive as rfc states. But as Oleksandr said, there's an issue w/ netty (and test is stuck). Will make changes on next release to be insensitive on header matching for captures in testing-http.

PR #971, opened 2022-11-28
fix(TDI-48957): TCK: stream-avro use "talend.studio.length" and "talend.studio.precision" to match the tck framework

wwang-talend:


pyzhou-talend (approved):
LGTM

PR #987, opened 2022-12-27
feat(TDI-48740): dssl support in substitutor

ypiel-talend:
https://jira.talendforge.org/browse/TDI-48740

### DIFF
--- a/common/src/main/java/org/talend/components/common/text/Substitutor.java
+++ b/common/src/main/java/org/talend/components/common/text/Substitutor.java
@@ -57,7 +57,7 @@ public class Substitutor {
             return source;
         }
 
-        int prefixLength = this.finder.getPrefix().length();
+        int prefixLength = this.finder.getPrefixWithKeyPrefix().length();
         int suffixLength = this.finder.getSuffix().length();
 
         if (source.length() < prefixLength + suffixLength) {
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/talend/components/common/text/Substitutor.java
```diff
@@ -13,206 +13,200 @@
...
- * possible.
- * 
- * this class doesn't treat embraced substitution '${word${number}}' isn't translate to 'third'
- * with a function that give '3' for 'number' and 'third' for 'word3'
- * 
- * it doesn't treat recursive replacement: '${greeting}' with a function
- * 'greeting' -> '${hello} ${world:-tdi}" and 'hello' -> 'hi' will render '${hello} ${world:-tdi}'.
- */
+
 public class Substitutor {
```
olhava, line 20: Would be nice to add a new javadoc instead

common/src/main/java/org/talend/components/common/text/Substitutor.java
```diff
@@ -13,206 +13,200 @@
...
+        private final String prefix;
 
-            if (result.start == 0 || source.charAt(result.start - 1) != ESCAPE) {
-                sb.append(source.substring(curr, result.start)).append(findOrDefault(result.key));
-                curr = result.end;
-            } else { // escaped placeholder
-                sb.append(source.substring(curr, result.start - 1));
-                curr = result.start;
-            }
+        private final String suffix;
```
olhava, line 154: I think suffix parameter could be renamed here to postfix. Or postfix and suffix could be renamed to open/closeSequence instead. prefixWithKeyPrefix can also benefit from such rename if it becomes openSequenceWithKeyPrefix.
Also KeyFinder class looks more like SubstitutorSettings.

ypiel-talend:
Modification in Substitutor can't be used in REG2 because, livy  has a transitive dependency on connectors-se/common thought common-dq, and so connector use common provided by livy and not the version they have in their dependency. Currently I remove this PR and move the new Substitutor in my -ee connector. Will be proposed in -se after once livy dependency issue will be fixed.

PR #996, opened 2023-01-19
feat(TDI-49274): bump component-runtime to 1.53.0

undx:
https://jira.talendforge.org/browse/TDI-49274


ozhelezniak-talend (approved):
lgtm

PR #997, opened 2023-01-20
fix(TDP-11708): dupplicate column on jdbc connector

jlame:
see TDP-11708.
When passing props to schema with different value, we can have the same column twice when different value are inside props for different records.
Solution : replace distinct with a distinct by key


ypiel-talend (approved):
LGTM

wwang-talend:
the fix remove all duplicated name columns for create database/insert/update sql and so on. yes, it can fix current issue, but we should ask those questions:

1. why tck record allow duplicated field name? it remember in tck record, it has name and rawName field, rawName for the origin name, name for tck tech name.
2. in business, for customer case, the duplicated columns are not valid always? impossbile customer's data struct have two columns which both is named : "date", maybe this is a seldom case, not sure.
3. seems the dataprep function add two columns : "date" and "id", that make the duplicated column thing, and seems we expect that two columns are not in target sink(here is database table), if that, imho, we should filter it before it reach jdbc sink connector.

The code is ok for me, no regression risk as it don't work before if duplicated column, but please consider the 3 questsions above.

jlame:
in fact data prep function doesn't add the column. 
The issue comes from the fact that data prep uses schema entry props to store some information on the record here it's the validity of the record.
The issue is that the connector tries to do a distinct on the schema entry to retrieve all columns to be created without dupplicate. The issue is that the distinct method is doing it based on the whole object and so takes into account the props inside the schema entry which are specific to TDP usage and shouldnt be taken into account in this distinct.
Maybe the name is not the one to be used ? i can change it to rawName if thats the case.

wwang-talend:
thanks @jlame 's explain, now understand it.

wwang-talend (approved):

PR #1009, opened 2023-03-16
fix(TDI-47593): support logical type

wwang-talend:


jdbc/src/main/java/org/talend/components/jdbc/output/platforms/MySQLPlatform.java
```diff
@@ -116,6 +116,30 @@ private String toDBType(final Column column) {
         case BYTES:
             return "BLOB";
         case DATETIME:
+            final String logicalTypeName = column.getProp(SchemaProperty.LOGICAL_TYPE);
+            log.info("logical type :" + logicalTypeName);
+            if (logicalTypeName != null) {
+                final LogicalType logicalType = LogicalType.valueOf(logicalTypeName);
+                final Integer scale = getInt(column, SchemaProperty.SCALE);
```
wwang-talend, line 123: should be int, will fix

PR #1013, opened 2023-03-21
feat(TDI-49459) Use maven apache checkstyle on connectors build

acatoire:
feat(TDI-49459) Use maven apache checkstyle on connectors build
https://jira.talendforge.org/browse/TDI-49459

The first need was to detect missing translations, but more checks can be done with checkstyle.
Improvement done on the ci build during this PR:
- sonar is on a separate stage now
- sonar use a bash script for easier maintenance
- uniformize scripts name starting with "mvn_"
- add maven log analysis on jenkins build report page


Jenkinsfile
```diff
@@ -116,10 +116,6 @@ pipeline {
         skipStagesAfterUnstable()
     }
 
-    triggers {
-        cron(env.BRANCH_NAME == "master" ? "@daily" : "")
```
ozhelezniak-talend, line 120: We don't have a trigger anymore?

Jenkinsfile
```diff
@@ -212,7 +216,7 @@ pipeline {
                             branch_description) = extract_branch_info("$env.BRANCH_NAME")
 
                             // Check only branch_user, because if there is an error all three params are empty.
-                            if (branch_user.equals("")) {
+                            if (branch_user == ("")) {
```
ozhelezniak-talend, line 219: Is "==" okay substitution for "equals"? I don't know about groovy, but in Java it's two different things, and for String objects we should use "equals"

Jenkinsfile
```diff
@@ -417,6 +421,58 @@ pipeline {
...
+                        """
+                    }
+                }
+            }
+            post {
+                always {
+                    script {
+                        println '====== Archive artifacts'
+                        println "Checkstyle report"
+                        archiveArtifacts artifacts: "**/checkstyle-result.xml", allowEmptyArchive: true, onlyIfSuccessful: false
```
ozhelezniak-talend, line 445: Is it okay to have empty archive?

ozhelezniak-talend (commented):
I left a few questions.

Jenkinsfile
acatoire: The trigger should be in flux now, not in the job, but I just saw that not, I will revert this, but it will need to be corrected

Jenkinsfile
acatoire: It means that if by any reason the file is not created, the build will not fail in this step, but if the file is not here, there is another error that will make the build falling

ozhelezniak-talend (approved):
LGTM, not an expert of build scripts and Jenkins configurations, but looks ok

PR #1021, opened 2023-03-27
fix(TDI-48413):change jdbc wrong label

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-48413


wwang-talend (commented):
seems the option name change and migration only for the model field name change, that for readable.

wwang-talend (approved):

PR #1034, opened 2023-04-06
chore(TDI-49637): Update json-smart for connectors-se

dmytro-sylaiev:


azure/pom.xml
```diff
@@ -47,6 +47,11 @@
         <artifactId>reactor-netty-http</artifactId>
         <version>1.0.25</version>
       </dependency>
+      <dependency>
+        <groupId>net.minidev</groupId>
+        <artifactId>json-smart</artifactId>
+        <version>2.4.9</version>
```
ypiel-talend, line 53: The cve is fixed in 2.4.10 as explained in veracode no ?
https://sca.analysiscenter.veracode.com/workspaces/X33hjMQ/issues/vulnerabilities/176795135

ypiel-talend (changes requested):
One comment

azure/pom.xml
dmytro-sylaiev: 2.4.9 also fixed the direct CVE about parsing nested jsons: https://github.com/netplex/json-smart-v2/releases/tag/2.4.9
https://mvnrepository.com/artifact/net.minidev/json-smart

2.4.10 only fixes additional bug about it, but 2.4.9 is used in daikon and I updated to 2.4.9 to be aligned with daikon

ypiel-talend (approved):
LGTM

PR #1038, opened 2023-04-11
fix(TDI-45302): Upgrade Delta Lake connectivity

wwang-talend:


coheigea:
@wwang-talend What is the status of this PR please?

PR #1039, opened 2023-04-11
fix(TDI-46121): require snowflake fields

ozhelezniak-talend:
* make snowflake connection fields required where it's relevant


undx (approved):

ozhelezniak-talend:
The issue was closed as can't fix.

PR #1042, opened 2023-04-14
chore(TCOMP-2407) Add snapshot server

acatoire:
Add Talend snapshot server to be able to retrieve talend development branches due to https://jira.talendforge.org/browse/TCOMP-2407


undx (approved):
lgtm

PR #1048, opened 2023-04-25
chore(TDI-49740) change sonar stage place in workflow

acatoire:
https://jira.talendforge.org/browse/TDI-49740
chore(TDI-49740) change sonar stage place in workflow

Correct simple import warning at the same time.


ypiel-talend (approved):
LGTM

PR #1051, opened 2023-05-02
feat(TDI-49325): migrate jenkins jobs to tsbi customBuilder

acatoire:
https://jira.talendforge.org/browse/TDI-49325
feat(TDI-49325) migrate to Jenkins jobs to tsbi customBuilder

Migration to the new image and minors improvements to align CI components.


.jenkins/jenkins_pod.yml
```diff
@@ -0,0 +1,34 @@
+
+apiVersion: v1
+kind: Pod
+spec:
+  imagePullSecrets:
+    - name: talend-registry
+  containers:
+    - name: main
+      # 3.0.9-20220930152032 is the last root version of custom-builder
```
quilicicf, line 9:
```suggestion
      # 3.0.9-20220930152032 is the last root version of custom-builder
      # If you upgrade, read the doc first: https://github.com/Talend/tsbi-images/blob/main/tsbi/custom-builder/README.md#migrate-from-any-builder-to-custom-builder-31
```

.jenkins/jenkins_pod.yml
```diff
@@ -0,0 +1,34 @@
...
+      volumeMounts: [
+        { name: efs-jenkins-connectors-se-m2, mountPath: /root/.m2/repository },
+        { name: efs-jenkins-connectors-asdf, mountPath: /root/.asdf/installs, subPath: installs },
+      ]
+      resources: { requests: { memory: 3G, cpu: '4' }, limits: { memory: 8G, cpu: '5' } }
+      env:
+        - name: DOCKER_HOST
+          value: tcp://localhost:2375
+    - name: docker-daemon
+      image: artifactory.datapwn.com/docker-io-remote/docker:19.03.1-dind
```
quilicicf, line 22: Docker 19 is dated :wink:

.jenkins/mvn_package.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -xe
+
+# This script allow to package the current project with maven on ci settings.xml
+# You can provide extra parameters to tune maven
+# $1: extra_mvn_parameters
+main() {
+  _EXTRA_MVN_PARAMETERS=("$@")
+  
+  mvn package --update-snapshots \
```
quilicicf, line 26: Using `--update-snapshots` on CI is a code smell (I'd even go as far as code stink :wink:).
This should at least be commented IMO. 
Makes your job run way too many HTTP requests, terrible for performance.

.jenkins/mvn_package.sh
```diff
@@ -0,0 +1,32 @@
...
+              --batch-mode \
+              --settings .jenkins/settings.xml \
+              ${_EXTRA_MVN_PARAMETERS[*]}
```
quilicicf, line 29: Don't know where you picked the convention of adding a leading `_` but I use it to signify that a construct comes from another file so I'm confused here :sweat_smile: 

Might be conflicting conventions.

.jenkins/jenkins_pod.yml
```diff
@@ -0,0 +1,34 @@
...
+          value: ""
+      securityContext:
+        privileged: true
+  volumes:
+    - name: efs-jenkins-connectors-se-m2
+      persistentVolumeClaim:
+        claimName: efs-jenkins-connectors-se-m2
+    - name: efs-jenkins-connectors-asdf
+      persistentVolumeClaim:
+        claimName: efs-jenkins-connectors-asdf
```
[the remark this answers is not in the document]
acatoire: corrected

.jenkins/mvn_package.sh
acatoire: Exact opposit for me!
From some bash convention to avoid confusion with environment var
We could align together

.jenkins/veracode-sca.sh
```diff
@@ -0,0 +1,49 @@
...
+_SRCCLR_API_TOKEN=${1}
+
+main() (
+  printf '##############################################\n'
+  printf 'Veracode SCA (Source Clear Analysis)\n'
+  printf '##############################################\n'
+
+  veracode_sca
+)
+
```
quilicicf, line 41:
```suggestion
main() (
  _SRCCLR_API_TOKEN="${1:?"$(usage 'Parameter "srcclr_api_token" is needed.')"}"

  printf '##############################################\n'
  printf 'Veracode SCA (Source Clear Analysis)\n'
  printf '##############################################\n'

  veracode_sca
)

function usage(){
  printf 'Run a Veracode SCA (Source Clear Analysis)\n'
  printf 'Usage : %s <srcclr_api_token>\n' "${0}"
  printf '\n'
  printf '%s\n' "${1}"
  printf '\n'
  exit 1
}
```

Rationale for the parameters changes:
* I don't understand how the current implementation can work. If the parameter is set, the condition should return an error code that's not caught by `&&` and it's running with `set -e` => it always exits immediately?
* The script parameters are passed to the `main` function as is, so checking the parameters inside `main` is not an issue
* This frees you to move the declaration of `usage` (see below)
* You can use the Bash parameter interpolation: `${1:?Message}` to check that a field is unset (`?`) or empty (additional `:`), and provide any message you want, even generated by another method. This is discussable... I find it's less of a pain to read but YMMV.

Rationale for moving usage:

* What your script really does should be easy to find (at the top). That's one of the reasons why we create the `main` function and call it at the end.
* Bash supports forward reference, meaning you can declare `usage` after it's used in `main`, as long as you __execute__ `main` after `usage` is defined

Jenkinsfile
```diff
@@ -247,7 +217,7 @@ pipeline {
                                 - Rename your branch
                                 """.stripIndent()
                                 currentBuild.description = ("ERROR: The branch name is not correct")
-                                sh """exit 1"""
+                                sh 'exit 1'
```
quilicicf, line 220: Woot?
Isn't it better to use `error("Failed because of a very good reason")`?

ci/Jenkinsfile-scan
quilicicf, line 1: It'd be nicer to change the naming convention of your Jenkinsfiles so that GitHub can do proper syntax coloration. Either:
* Only call them `Jenkinsfile` (then obviously put them in folders whose names explain what the Jenkinsfile does)
* Add the suffix `.groovy`

ci/Jenkinsfile-scan
```diff
@@ -1,52 +1,32 @@
...
-        - name: talend-registry
-""".stripIndent()
+// Maven package skipped elements
+// Everything that is not needed for the veracode scan is excluded to reduce time
+_SKIP_OPTS = "--define maven.test.skip=true\
+              --define spotless.apply.skip=true\
+              --define checkstyle.skip=true\
+              --define rat.skip=true\
+              --define skipTests=true\
+              --define invoker.skip=true"
```
quilicicf, line 17: This will generate a string with a lot of white space which can clutter the logs.
I like to build these strings like this:

```suggestion
_SKIP_OPTS = [
  '--define maven.test.skip=true',
  '--define spotless.apply.skip=true',
  '--define checkstyle.skip=true',
  '--define rat.skip=true',
  '--define skipTests=true',
  '--define invoker.skip=true'
].join(' ')
```

YMMV

acatoire: I like your solution very much!

ci/Jenkinsfile-scan
acatoire: I agree but, for now it is like this in every connectors' repos, I may do a ticket to re-align this later

.jenkins/jenkins_pod.yml
acatoire: What should be the rule to update it?
Historically in our team we never touch it, I think

quilicicf: The rule in my team is, if you witness an out-of-date lib (whatever it is), you try to update it and
* If it still works, good job
* If not, you revert an open an issue to be taken for next iteration

It's served us well along the years :)

.jenkins/mvn_package.sh
quilicicf: Open an issue and get to it ASAP, friend's advice.
Snapshots are the cause of 70% of your suffering on CI IMO.

.jenkins/mvn_package.sh
quilicicf: Yeah, would definitely be nice to align on this :sweat_smile: 
Personally I do:
* SCREAMING_SNAKE_CASE => only for environment or global constants
* camelCase => for variables and functions
* _SNAKE_CASE_orCamelCaseWithLeadingUnderscore => for anything that comes from another file

The reason I'm using additional characters for outside constructs is that it can also apply to methods so all cases are unambiguous.

ci/Jenkinsfile-scan
quilicicf: As you wish, just be aware that it's real hard on the reviewers :sweat:

.jenkins/jenkins_pod.yml
acatoire: seem good

ci/Jenkinsfile-scan
acatoire: I will make a ticket on that

Jenkinsfile
```diff
@@ -322,6 +299,7 @@ pipeline {
                         """
                     }
 
+                    sleep(time:3, unit:"SECONDS") // Ensure the docker container is fully up and running.
```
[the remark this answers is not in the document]
acatoire: Rertry version committed and  working on ci

quilicicf (approved):

quilicicf (approved):

PR #1052, opened 2023-05-03
feat(TDI-49769): remove connectors-se-docker module

undx:
https://jira.talendforge.org/browse/TDI-49769


ozhelezniak-talend (approved):
lgtm

PR #1059, opened 2023-05-09
feat(TDI-49786):upgrade mysql driver.

vyu-talend:
https://jira.talendforge.org/browse/TDI-49786


jdbc/src/main/java/org/talend/components/jdbc/service/JdbcService.java
```diff
@@ -587,9 +590,12 @@ public void addColumn(final Record.Builder builder, final Schema.Entry entry, fi
             }
             break;
         case DATETIME:
-            builder
-                    .withDateTime(entry,
-                            value == null ? null : new Date(((java.util.Date) value).getTime()));
+            if (value instanceof LocalDateTime) {
```
pyzhou-talend, line 593: If value == null then NPE here

pyzhou-talend, line 593: if value == null, it will work, but the logic is quite fishy
value instanceof LocalDataTime return false
Then the else will take care of null value
please change the logic

pyzhou-talend (approved):
LGTM

PR #1063, opened 2023-05-11
fix(TDI-49610): set empty string as null for guid in dynamics-crm

mhryb:
jira ticket: https://jira.talendforge.org/browse/TDI-49610


dynamics-crm/src/main/java/org/talend/components/dynamicscrm/output/AbstractToEntityRecordProcessor.java
```diff
@@ -335,7 +336,7 @@ private ClientProperty processString(Record fromRecord, EdmElement propType, Str
             EdmPrimitiveTypeKind primitiveTypeKind =
                     EdmPrimitiveTypeKind.valueOfFQN(propType.getType().getFullQualifiedName());
             if (primitiveTypeKind == EdmPrimitiveTypeKind.Guid) {
-                if (value == null) {
+                if (value == null || StringUtils.isEmpty(value)) {
```
dmytro-sylaiev, line 339: `public static boolean isEmpty(final CharSequence cs) {
        return cs == null || cs.length() == 0;
    }`

no need to explicitly check for nulls then

dmytro-sylaiev (approved):
LGTM

mhryb:
jira ticket: https://jira.talendforge.org/browse/TDI-49610

PR #1072, opened 2023-05-25
feat(TDI-49791):upgrade salesfroce driver and support latest endpoint.

vyu-talend:
https://jira.talendforge.org/browse/TDI-49791


pyzhou-talend (approved):
LGTM

PR #1078, opened 2023-06-02
feat(TDI-49728):upgrade snowflake driver for cve.

vyu-talend:
https://jira.talendforge.org/browse/TDI-49728


jzhao-talend (approved):
LGTM

PR #1079, opened 2023-06-06
fix(TDI-49870): Fix adlsInput ignores header value

dmytro-sylaiev:


azure/azurecommon/src/main/java/org/talend/components/common/formats/csv/CSVFormatOptions.java
```diff
@@ -18,6 +18,7 @@
 import org.talend.components.common.formats.Encoding;
 import org.talend.sdk.component.api.configuration.Option;
 import org.talend.sdk.component.api.configuration.condition.ActiveIf;
+import org.talend.sdk.component.api.configuration.ui.DefaultValue;
```
mhryb, line 21: Seems like unnecessary import

mhryb (approved):

azure/azurecommon/src/main/java/org/talend/components/common/formats/csv/CSVFormatOptions.java
dmytro-sylaiev: 👍 thank you for noticing, removed

mhryb (approved):

mhryb (approved):

mhryb (approved):

PR #1086, opened 2023-06-27
 chore(TDI-49964): Remove next snapshot build action

acatoire:
https://jira.talendforge.org/browse/TDI-49964
Remove the post release build
Extend timeout value because the build is close to 1h


fdesiles (approved):
LGTM

PR #1130, opened 2023-10-20
feat(TDI-49775): add support of application default credentials

mhryb:


dmytro-sylaiev (dismissed):
LGTM but also please add `@DefaultValue` for enum

dmytro-sylaiev (approved):

PR #1133, opened 2023-10-25
chore(TDI-50424): Extract proxy config

dmytro-sylaiev:
* Add migration

* Fix tests


http-common/src/main/java/org/talend/components/http/migration/HttpClientRuntimeMigrationHandler.java
```diff
@@ -0,0 +1,31 @@
...
+import org.talend.sdk.component.api.component.MigrationHandler;
+
+import static org.talend.components.http.migration.HttpClientDatastoreMigrationHandler.migrateProxyConfig;
+
+public class HttpClientRuntimeMigrationHandler implements MigrationHandler {
+
+    @Override
+    public Map<String, String> migrate(int incomingVersion, Map<String, String> incomingData) {
+        if (incomingVersion < 2) {
+            migrateProxyConfig(incomingData, "configuration.dataset.datastore.");
```
ypiel-talend, line 26: Maybe rename to "migrateProxyConfig_1_to_2" since we could have proxy migration in another future version ?

mhryb (approved):

ypiel-talend (approved):
LGTM
Please test migration part with a cloud pipeline pushed in TMC.

http-common/src/main/java/org/talend/components/http/migration/HttpClientRuntimeMigrationHandler.java
dmytro-sylaiev: in that case, package-private method might be renamed I think if needed (or add another method with some specific name), I'm good with the current name

mhryb (approved):

PR #1134, opened 2023-10-25
chore(TFD-16346 TFD-16134): fix CVEs about snappy-java and snakeyaml

amaumont:


pyzhou-talend:
Closed as the fix is done in https://jira.talendforge.org/browse/TDI-50450

PR #1137, opened 2023-11-01
fix(TDI-50461): [JDBC TCK]: remove the unuseful tjdbctableinput connector which is generated as we define the @Emitter class for cloud platform

wwang-talend:


jzhao-talend (approved):
LGTM

wyuan (approved):

PR #1157, opened 2023-12-06
chore(TDI-49473): Dependency convergence not used on CI build

acatoire:
Dependency convergence not used on CI build but on dev branch build
* Use property to select light rules set

https://jira.talendforge.org/browse/TDI-49473


ozhelezniak-talend (approved):
lgtm
looks okay, it'd like to see how it works in practice
from pom perspective I don't have objections

ozhelezniak-talend (approved):
lgtm

PR #1162, opened 2023-12-13
fix(TDI-50573): Connection pool thread not killed on component-server

wwang-talend:


ozhelezniak-talend (dismissed):
lgtm
I observed a couple times that a classes of the connector were loaded not by ConfigurableClassLoader of the connector, but the parent one. Sometimes it's classes of plugin, sometimes we just want to load a new classes with the configurableclass loader of the connector. 
That's why in some connectors we have the hack: invokeInLoader. Where we swap the thread context classloader and invoke the code with the ConfigurableClassloader.

Not 100% sure that it's related. But the class loading of the connectors a bit flaky and not the same in the different environments.

acatoire (dismissed):
Is it possible to correct the sonar?

jzhao-talend (approved):

PR #1164, opened 2023-12-13
fix(TDI-50640): Fix file upload

dmytro-sylaiev:


http-common/src/main/java/org/talend/components/http/service/httpClient/HTTPClientService.java
```diff
@@ -237,17 +238,21 @@ public QueryConfiguration convertConfiguration(final RequestConfig config, final
...
+                contentDispositionString.append("; filename=\"").append(file.getName()).append("\"");
 
-                List<String> contentDispositionList = new ArrayList<>();
-                contentDispositionList.add("attachment");
-                contentDispositionList.add("file=\"" + uploadFile.getName() + "\"");
-                contentDispositionList.add("filename=\"" + file.getName() + "\"");
                 try {
-                    contentDispositionList.add(
-                            "filename*=UTF-8''\"" + URLEncoder.encode(file.getName(), "UTF-8") + "\"");
+                    contentDispositionString.append("; filename*=UTF-8''\"")
```
ypiel-talend, line 248: Is that normal to have double ' ? filename*=UTF-8''

ypiel-talend (commented):
1 question

http-common/src/main/java/org/talend/components/http/service/httpClient/HTTPClientService.java
dmytro-sylaiev: > This example is the same as the one above, but adding the "filename"
>    parameter for compatibility with user agents not implementing
>    [RFC 5987](https://www.rfc-editor.org/rfc/rfc5987):
> 
>      Content-Disposition: attachment;
>                           filename="EURO rates";
>                           filename*=utf-8''%e2%82%ac%20rates

https://www.rfc-editor.org/rfc/rfc6266#section-6

ypiel-talend (approved):
LGTM

PR #1177, opened 2023-12-29
chore(TDI-50717): Fix double junit result in TTO

acatoire:
Remove JUnit plugin as the publishing for TTO is already done by realtimeJUnit.
Add test to artifacts as it is a good practice regarding test analysis.

https://jira.talendforge.org/browse/TDI-50717


fdesiles (approved):
LGTM

